PR #75, opened 2016-10-05
Fix business changes:

duker33:
- Change header cart animation config
- Fix ftp connection for import catalog
- Autolaunch collectstatic on catalog command
- Fix price excel markup


shopelectro/management/commands/catalog.py
```diff
@@ -271,9 +279,18 @@ def remove_xml(self) -> result_message:
     @process('Create price lists')
     def generate_prices() -> result_message:
         """Generate Excel, YM and Price.ru price files."""
-        commands = ['excel', 'price']
+        commands = [
```
wrtds2: Можно структуру сделать такой:
```
commands = [
    ('excel', ),
    ('price', ),
    ('collectstatic', '--noinput')
]
```
и 
```
for command in commands:
    call_command(*command)
```

PR #133, opened 2017-04-19
Hide orphan prods

dveselov:
* https://trello.com/c/9BbCywDe/242-se-hide-orphan-prods

### DIFF
--- a/shopelectro/tests/tests_views.py
+++ b/shopelectro/tests/tests_views.py
@@ -162,16 +162,12 @@ class YandexKassa(TestCase):
 
 class ProductPage(TestCase):
 
-    def test_orphan_product(self):
-        product = Product.objects.create(
-            name='Orphan product',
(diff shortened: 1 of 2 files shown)


shopelectro/tests/tests_views.py
```diff
@@ -160,6 +160,21 @@ def test_yandex_aviso_body(self):
         self.assertContains(response, 'code="1"')
 
 
+class ProductPage(TestCase):
+
+    def test_orphan_product(self):
+        product = Product.objects.create(
```
artemiyds2: Лучше заюзай [фикстуры](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#fixture-loading)

artemiyds2:
```
```
product = Product.objects.first()
response = self.client.get(product.url)
....
```

PR #271, opened 2018-03-22
#fix  Add link on team to README

duker33:


duker33:
CI layed down, when i created PR.
I fixed only README, so i merge it without CI

Issue #250, opened 2017-12-26
docker-compose.yml:56-57: if Validate `.env` is full Is...


0pdd:
The puzzle `229-a984cd8e` from #229 has to be resolved:

https://github.com/fidals/shopelectro/blob/5308166c782497f51ba645587af67377e152720b/docker/docker-compose.yml#L56-L57

The puzzle was created by duker33 on 26-Dec-17. 

Estimate: 0 minutes, role: IMP.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@artemiyds2 @duker33 please pay attention to this new issue.

duker33:
connected with #280

PR #280, opened 2018-04-01
#269 Refactor deploy system

### LINKED ISSUE #269: Улучшить билд систему
Зависит от #248 

----

Как показала практика, для новых разработчиков вникнуть в систему сборки оказалось одной из самых больших задач. Поэтому мы должны упростить ее.
- билдим образы только на CI -> пушим их на docker-hub -> разработчики пулят их у себя на компах. Это поможет избежать проблем со сборкой образов.
- задокументировать систему сборки. Если кому-то придется разбираться в ней, то это будет проще с документацией
- переименовать .env.dist в .env, чтобы не делать лишней работы
- монтировать контейнеры только внутрь проекта. Мы не должны монтировать ничего в корень системы (e.g. `/opt/database/shopelectro:/var/lib/postgresql/data` -> `../database/postgres:/var/lib/postgresql/data`). P.S. Актуально только для `docker-compose.yml`, т.е. для локальной разработки

В идеале процесс доставки зависимостей на компьютер разработчика будет таким:
```bash
docker-compose pull
docker-compose up
```

artemiy312:
Closes #269


docker/Makefile
```diff
@@ -1,60 +1,72 @@
...
+	$(dc) up -d app selenium
+	$(dc) exec app python manage.py test -v 3 --liveserver=app:8020-8030 --parallel
 	$(dc) stop
 
+lint:
+	$(dc) up -d app
+	$(dc) run --rm lint
+	$(dc) stop
+
+deploy-dev:
```
duker33: let's add to this rule comment like this:
```
# depends on `build-static`. Don't set this dependency explicitly for caching purposes
```

docker/build/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+  python-dev:
+    build:
+      context: ../../
+      dockerfile: docker/images/python/dev/Dockerfile
+    image: fidals/se:dev
+
+  python-prod:
+    build:
+      context: ../../
+      dockerfile: docker/images/python/prod/Dockerfile
```
duker33: maybe we can split dockerfiles themself instead of subdirectoring:
`docker/images/python/Dockerfile.(prod|dev)`

docker/docker-compose.yml
```diff
@@ -2,103 +2,72 @@ version: '2'
...
-      - REDIS_LOCATION_THUMBNAIL=redis://se-redis:6379/2
-      - REDIS_LOCATION_USER_AGENT=redis://se-redis:6379/3
-      - FTP_USER=$FTP_USER
-      - FTP_PASS=$FTP_PASS
-      - FTP_IP=$FTP_IP
-      - SECRET_KEY=$SECRET_KEY
-      - YANDEX_SHOP_PASS=$YANDEX_SHOP_PASS
-      - EMAIL_HOST_PASSWORD=$EMAIL_HOST_PASSWORD
-      - ENV_TYPE=$ENV_TYPE
+      - REDIS_LOCATION_DEFAULT=redis://redis:6379/0
```
duker33: let's put it to some env file. Maybe seprated: `env_files/urls` / `env_files/service_urls` or smth.

Now it's duplicated in compose files

docker/create-env.sh
```diff
@@ -0,0 +1,62 @@
...
+        then
+            echo "Checkout the $new_file and configure it."
+            read -p "Are you ready to continue? [y/n]: " yn
+            if [[ $yn != "y" ]]
+            then
+                exit 1
+            fi
+        fi
+
+        cat $new_file >> .env
```
duker33: wait ex dist files modification is not obvious and clear for use.
We should explain several steps throw bash script interface or throw doc.
Moreover `.env` file contains constants duplication

Only `build/docker-compose.yml` uses `.env` file, as i see.

We can just add smth like it to this docker-compose:
```
env_file:
 - env_files/app
 - env_files/credentials
 - env_files/paths
 - env_files/ports
```

... and rm `.env` file with it's additional bash dialog

duker33:
@artemiy312 , tried to test.

create-env is the first rule, i launched. Got this result

```
andre@debian:~/prog/shopelectro/docker$ m create-env
Are you ready to continue? [y/n]: y
Are you ready to continue? [y/n]: y
Are you ready to continue? [y/n]: y
Are you ready to continue? [y/n]: y
Checkout the env_files/app and configure it. Checkout the env_files/credentials and configure it. Checkout the env_files/paths and configure it. Checkout the env_files/ports and configure it.
make: Checkout: Команда не найдена
Makefile:15: ошибка выполнения рецепта для цели «create-env»
make: *** [create-env] Ошибка 127
```

artemiy312:
@duker33 

I tried to run it in debian container and it looks ok.

```bash
root@e8287ef3bc02:/src# rm env_files/ports env_files/credentials  env_files/paths env_files/app .env
root@e8287ef3bc02:/src# ./create-env.sh
Checkout the env_files/app and configure it.
Are you ready to continue? [y/n]: y
Checkout the env_files/credentials and configure it.
Are you ready to continue? [y/n]: y
Checkout the env_files/paths and configure it.
Are you ready to continue? [y/n]: y
Checkout the env_files/ports and configure it.
Are you ready to continue? [y/n]: y
root@e8287ef3bc02:/src# ls -lh .env
-rw-r--r-- 1 root root 1.2K May  5 23:48 .env
root@e8287ef3bc02:/src# rm env_files/ports env_files/credentials  env_files/paths env_files/app .env
root@e8287ef3bc02:/src# ls -lh .env
ls: cannot access '.env': No such file or directory
root@e8287ef3bc02:/src# ./create-env.sh --quite
root@e8287ef3bc02:/src# ls -lh .env
-rw-r--r-- 1 root root 1.2K May  5 23:48 .env
```

Can you execute it directly from console?

docker/create-env.sh
artemiy312: @duker33 
>wait ex dist files modification

We do not wait dist files modification, instead we copy dist files and wait modifications of these copies. `create-env.sh` already has doc for this process. 

>Only build/docker-compose.yml uses .env file, as i see.

The `.env` file is required to configure options of docker-compose files such as: `$SRC_DIR`, `$DEPS_DIR`, `$REDIS_PASSWORD`, etc.
The `env_file` option pass env variables only to container, but not to `docker-compose.yml`

docker/build/docker-compose.yml
artemiy312: We can, but what benefits it will give us?

duker33: @artemiy312 , we'll avoid unnecessary subdirectoring. It seems enough benefit for me =)

docker/create-env.sh
duker33: @artemiy312 , 

> We do not wait dist files modification, instead we copy dist files and wait modifications of these copies.

y, copies, but it does not matter for UX. I thought about it for a little. We have not simple UX decision only because of not simple task. So, not UX is complex, but designed good. 
Only one moment: i think we can copy all files the same time. And then ask user to edit them in a bunch.
He'll push only one "y" button instead of several

> The env_file option pass env variables only to container

got it. Another non-obvious docker-composer detail. One not good thing, that stayed: after `create-env.sh` launched we'll have `.env` file in repo root and `env_files/` too. It'll be confusing.
Let's name it `.env.build` and add this comment `Used only for images build. See env_files/ folder to change containers behaviour.`

duker33:
@artemiy312 , with bash it works fine:

Makefile works good after this modification:
```
create-env:
	@bash ./create-env.sh
```

docker/create-env.sh
artemiy312: @duker33 
>And then ask user to edit them in a bunch.

I going to do it

>after create-env.sh launched we'll have .env file in repo root and env_files/ too. It'll be confusing.

I do not get the point. Currently,  the `create-env.sh` creates only one `.env` file in `docker/` directory, isn't it? Also we can not to name an environment file as `.env.build`, because docker-compose requires certain name for it - `.env`

duker33: @artemiy312 , 

> Currently, the create-env.sh creates only one .env file in docker/ directory, isn't it?

but we also have `env_files/` dir with vars. It's not obvious, that .env is only for build, but env_files/ are only for containers. Every new proger (or we after half a year) will spend time on debugging and config digging trying to change env var for container

> Also we can not to name an environment file as .env.build

hmm, it's strange, `.env` file is visible for build, `env_file: .env.build` is not visible already.
If it's another docker-compose amazing detail, then we have to leave it as is. But we can add comment at top of this file

docker/docker-compose.yml
duker33: @artemiy312 , what about this comment? Review fix doesn't cover it

duker33:
@artemiy312 , another error:
```
andre@debian:~/prog/shopelectro/docker$ m build-static
docker-compose run --rm nodejs gulp build
Creating docker_app_1
Pulling nodejs (fidals/se-nodejs:prod)...
```

fix pls `image: fidals/se-nodejs:prod` to `image: fidals/se-nodejs:dev`.
We use nodejs only for dev purposes, and have it's only `dev` tag in docker repo

docker/create-env.sh
duker33: tested it:

```
# docker/images/python/dev/Dockerfile
FROM python:slim

RUN echo $SRC_DIR && echo 'bugaga'
...
```

got this output:
```
andre@debian:~/prog/shopelectro/docker$ m create-env
# many "y"
andre@debian:~/prog/shopelectro/docker$ dc -f build/docker-compose.yml build --no-cache python-dev
Building python-dev
Step 1/6 : FROM python:slim
 ---> d6f22b3a2b87
Step 2/6 : RUN echo $SRC_DIR && echo 'bugaga'
 ---> Running in 550af64346e3

bugaga
...
```
It seems build accept no env files, even `.env` in context root. And dc doc says, that it is:
https://docs.docker.com/compose/compose-file/compose-file-v2/#env_file

And every our dockerfile doesn't use any var from .env.

So, it seems we can just rm `.env` file with it's creation in bash code.

duker33: @artemiy312 , this one was not fixed:

> He'll push only one "y" button instead of several

artemiy312: @duker33 

>So, it seems we can just rm .env file with it's creation in bash code.

`.env` file is required neither for building images nor substitution in `Dockerfile`, but for `docker-compose.yml` execution. We use substitution in  `docker-compose.yml`

```bash
rm .env
docker-compose up python
WARNING: The VIRTUAL_HOST_EXPOSE_PORT variable is not set. Defaulting to a blank string.
WARNING: The VIRTUAL_HOST_PORT variable is not set. Defaulting to a blank string.
WARNING: The VIRTUAL_HOST_LIVESERVER_PORT variable is not set. Defaulting to a blank string.
WARNING: The SRC_DIR variable is not set. Defaulting to a blank string.
WARNING: The DEPS_DIR variable is not set. Defaulting to a blank string.
WARNING: The REDIS_PASSWORD variable is not set. Defaulting to a blank string.
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.app.ports is invalid: Invalid port "", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]
services.app.ports is invalid: Invalid port ":", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]
``` 

>this one was not fixed:

yes, because I still do not push any changes

docker/Makefile
artemiy312: @duker33 

`se-nodejs:dev` already has built static files, so will be enough to pull it.

docker/create-env.sh
duker33: @artemiy312 

> but for docker-compose.yml execution

but `dc build` works fine. And it's single purpose of build/docker-compose.yml file, as i see.
It should not up any containers

> yes, because I still do not push any changes

i thought, i commented about it at the first review, sorry.

artemiy312: @duker33 

`.env` file is required for `docker/docker-compose*.yml` files

duker33:
@artemiy312 , plz take care of connected issues https://github.com/fidals/refarm-site/issues/94 and https://github.com/fidals/shopelectro/issues/250

Update them according to this PR

docker/create-env.sh
artemiy312: @duker33 You can try to execute `docker-compose up`  without `.env` and you will see a such error: https://github.com/fidals/shopelectro/pull/280#discussion_r186714135.

As I said before, the `.env` is required for variable substitution in `docker/docker-compose*.yaml`

e.g.

```yaml
  app:
    image: fidals/se:prod
    container_name: se-python
    restart: always
    env_file:
      - ...
    environment:
      - ...
    depends_on:
      - ...
    ports:
      - $VIRTUAL_HOST_PORT  # it can only be taken from .env
```

docker/docker-compose-production.yml
```diff
@@ -2,245 +2,170 @@ version: '2'
...
-  se-celery-beat:
+  celery-beat:
     image: fidals/se:prod
     container_name: se-celery-beat
     restart: always
+    env_file:
+      - env_files/app
+      - env_files/credentials
+      - env_files/paths
     environment:
```
duker33: @artemiy312 , `environment` key contain no values, docker-compose throws error because of it

docker/create-env.sh
duker33: @artemiy312 , tested manually. You are right.
Here is good explanation
https://github.com/docker/compose/issues/4223#issuecomment-280077263

duker33:
added two comments to fix:
- https://github.com/fidals/shopelectro/pull/280#discussion_r187795154
- https://github.com/fidals/shopelectro/pull/280#discussion_r187793888

docker/create-env.sh
duker33: fixed

artemiy312:
@duker33 I going to rebase the PR to test code with new changes. After this we may merge

duker33:
@artemiy312 , check plz few little fixes here https://github.com/fidals/shopelectro/pull/280/commits/58c7de39cc17c6c21c6dc36ae92d0355a3b62cf2

PR #293, opened 2018-05-12
#190 Create backup service

### LINKED ISSUE #190: Настрой систему резервного копирования
Depends on #234 

------

- [x] Создавай бэкап-скрипты
- [ ] Запускай их на прод-серве по крону

artemiy312:
Closes #190


etc/crontab
```diff
@@ -0,0 +1,3 @@
+# make backups every monday, wednesday and friday at 23:00
```
duker33: let's move all cron staff to `backup` container. In this way every service will have it's own inner schedule.
Host machine should know nothing about it

If we go this way:
- we can bind backup script to fs events instead of cron in the future
- we can simply reuse the same backup image and docker service with different sites

etc/se-backup-entrypoint.sh
```diff
@@ -6,5 +6,8 @@
 for type in database media static
```
duker33: we can move `etc/se-backup-*.sh` to `etc/backup`

PR #306, opened 2018-05-19
#244 Add explicit wait of element to fix test

### LINKED ISSUE #244: tests_selenium.py:1103-1105: <body>Fix a...
The puzzle `SEARCH-TEST-1f4dde45` from #SEARCH-TEST has to be resolved:

https://github.com/fidals/shopelectro/blob/013da6bd56f0a3e15309e9a4343c679c8a28aae4/shopelectro/tests/tests_selenium.py#L1103-L1105

The puzzle was created by Artemiy on 20-Dec-17. 

Estimate: 0 minutes, role: IMP.

artemiy312:
Closes #244

### DIFF
--- a/shopelectro/tests/tests_selenium.py
+++ b/shopelectro/tests/tests_selenium.py
@@ -1052,7 +1052,7 @@ class Search(helpers.SeleniumTestCase):
         self.browser.find_element_by_class_name('search-form').submit()
         self.wait.until(EC.url_contains('/search/'))
         self.wait.until(EC.visibility_of_element_located(
-            (By.TAG_NAME, 'H1')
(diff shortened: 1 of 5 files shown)


shopelectro/tests/tests_selenium.py
```diff
@@ -1051,6 +1051,9 @@ def clear_input(self):
     def search(self):
         self.browser.find_element_by_class_name('search-form').submit()
         self.wait.until(EC.url_contains('/search/'))
+        self.wait.until(EC.visibility_of_element_located(
+            (By.TAG_NAME, 'H1')
```
duker33: `h1` (lower case) is common way to identify tags

PR #312, opened 2018-05-30
#309 Reduce retail prices. Create docs for PRICE_REDUCER const.

### LINKED ISSUE #309: Price round problem: product page and cart
Before this moment we had requirement to do `-1%` for price fetched from 1C.

But we round this price on product page. Look at the screen: two products cost less then `price x 2` because of this UI rounding.

To solve it PO created new requirement: don't do `-1%` for price, but do `-1 руб` for every retail (not wholesale) price.
Do it in fetching data from 1C

artemiy312:
Closes #309

### DIFF
--- a/shopelectro/settings/base.py
+++ b/shopelectro/settings/base.py
@@ -293,9 +293,8 @@ INTERNAL_IPS = (
 
 TOP_PRODUCTS = [291, 438, 1137, 2166, 2725, 2838, 3288, 3884, 3959, 2764]
 
-# reduce product prices by PRICE_REDUCER to divide
-# the retail-related site (shopelector) from the wholesale-related site (se78)
-# in web search engines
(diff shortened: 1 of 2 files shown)


shopelectro/settings/base.py
```diff
@@ -293,7 +293,10 @@
 
 TOP_PRODUCTS = [291, 438, 1137, 2166, 2725, 2838, 3288, 3884, 3959, 2764]
 
-PRICE_MULTIPLIER = 0.99
+# reduce product prices by PRICE_REDUCER to divide
+# the retail-related site (shopelector) from the wholesale-related site (se78)
```
duker33: suggest to restyle comment:
```
decrement every product's price for PRICE_REDUCER value (in rub).
It's required to make prices on shopelectro.ru and se78.ru unique.
```

PR #316, opened 2018-05-31
#308 Remove duplications from search results.

### LINKED ISSUE #308: Page-product rows duplication on search page
image

artemiy312:
Closes #308


shopelectro/models.py
```diff
@@ -295,3 +296,17 @@ def serialize_tags_to_title(tags: TagQuerySet) -> str:
...
     )
+
+
+class PageExcludeModelTManager(TreeManager):
+
+    def get_queryset(self):
+        return super().get_queryset().exclude(type=Page.MODEL_TYPE)
+
+
+class PageExcludeModelT(Page):
```
duker33: `PageModelExcludedTManager` will be more clear.
Page don't exclude Model here. "Page" and "Model" should be together. `PageModel` is one entity

shopelectro/tests/tests_views.py
```diff
@@ -395,3 +398,32 @@ def test_admin_autocomplete_no_results(self):
...
+            response.content.decode('utf-8'), 'html.parser'
+        ).find_all(class_='search-result-link')
+        self.assertTrue(len(result_links) == 1)
+
+    def test_autocomplete_has_no_model_pages(self):
+        """Autocomplete does not contain page with type=MODEL_TYPE and duplicated content."""
+        response = self.client.get(
+            f'{reverse("autocomplete")}?term={self.QUOTED_SIGNLE_RESULT_TERM}'
+        )
+        result = list(filter(lambda d: d['type'] == 'category', json_to_dict(response)))
```
duker33: we check, that we have pretty one page with type category.
To check, that autocomplete has no model pages, we should filter by category and compare result's size with 0.
And admin_autocomplete (test below) too

shopelectro/models.py
artemiy312: @duker33 
>Page don't exclude Model here

I meant here Page model without ["MODEL_TYPE"](https://github.com/fidals/refarm-site/blob/master/pages/models.py#L74) pages, so maybe `ExcludedModelTPage` will be more appropriate?

duker33: @artemiy312 , got you point. It might be good to start name with Page, but it maybe impossible.
So, choose name for your own plz :)

You may find interesting some variants:
`NotModelPage`, `PageWithNoModelT`, `PageWithoutModelT`

shopelectro/tests/tests_views.py
```diff
@@ -415,9 +415,11 @@ def test_autocomplete_has_no_model_pages(self):
         response = self.client.get(
             f'{reverse("autocomplete")}?term={self.QUOTED_SIGNLE_RESULT_TERM}'
         )
-        result = list(filter(lambda d: d['type'] == 'category', json_to_dict(response)))
-        self.assertTrue(
-            len(result) == 1
+        self.assertFalse(
+            list(filter(
+                lambda r: r['type'] != 'category' and r['type'] != 'see_all',
```
duker33: `r['type'] not in ['category', 'see_all']`.
It'll be good to compare with type 'page' (or 'page_model') explicitly. It's technically impossible?

PR #324, opened 2018-06-03
#313  Create prepare-deploy rule. Recreate nginx container during deploy

### LINKED ISSUE #313: Deploy mech broken
`make deploy` leads to 500 error on prod because of error in `front/build` code shipping

```
# dcp logs app
se-python                | ValueError: The file 'js/index.min.js' could not be found with <django.contrib.staticfiles.storage.ManifestStaticFilesStorage object at 0x7ff92563ce80>.
```

duker33:
Closes #313

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -98,8 +98,8 @@ prepare-deploy:
 deploy:
 	$(dcp) pull
 	$(dcp) stop
-	$(dcp) rm -f app && $(dcp) up -d app
-	$(dcp) rm -f nginx && $(dcp) up -d nginx
+	$(dcp) rm -f app nginx
+	$(dcp) up -d app nginx
 	$(dcp) exec app python manage.py migrate
 	$(MAKE) -j generate-production-static-data
 	# Launch "collectstatic" not in static recipe because ManifestStaticStorage writes to db
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -84,20 +84,22 @@ generate-production-static-data:
 	$(dcp) exec app python manage.py excel
 	$(dcp) exec app python manage.py price
 
+# drone should do this in working flow.
+# But if drone don't do this for some reasons,
+# you can run this recipe from local machine.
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -64,11 +68,22 @@ generate-production-static-data:
...
+	$(dcb) build --no-cache python-dev python-prod
+	$(dcb) push python-dev python-prod
+
 deploy:
 	$(dcp) pull
 	$(dcp) stop
-	$(dcp) rm -f app
-	$(dcp) up -d app
+	$(dcp) rm -f app && $(dcp) up -d app
+	$(dcp) rm -f nginx && $(dcp) up -d nginx
```
artemiy312:
```bash
$(dcp) rm -f app  nginx && $(dcp) up -d app nginx
# or
# I'd prefer the second approach
$(dcp) rm -f app nginx
$(dcp) up -d app nginx
```
will be better

docker/Makefile
```diff
@@ -64,11 +68,22 @@ generate-production-static-data:
 	$(dcp) exec app python manage.py excel
 	$(dcp) exec app python manage.py price
 
+# drone should do this in working flow.
+# But if drone don't do this for some reasons,
+# you can run this recipe from local machine.
+prepare-deploy:
```
artemiy312: Seems `deploy-(preprare|push)` will be better. Decide on your own

docker/Makefile
```diff
@@ -64,11 +68,22 @@ generate-production-static-data:
...
+	rm -rf front/build
+	$(dc) stop app && $(dcp) stop app
+	$(dc) rm -f app && $(dcp) rm -f app
```
artemiy312: 1. Why do you remove an `app` container?
2. You may use here `docker-compose rm -fs ...` to stop and remove container

docker/Makefile
duker33: TODO

PR #329, opened 2018-06-05
stb167  Refactor config

duker33:
https://github.com/fidals/stroyprombeton/issues/167


docker/create-env.sh
```diff
@@ -24,23 +24,29 @@ function join_by { local IFS="$1"; shift; echo "$*"; }
...
         then
             cp $file $new_file
         fi
+    }
+
+    for file in env_files/*.dist
+    do
+        new_file $file
     done
+    new_file ../shopelectro/settings/local.py.dist
```
artemiy312: The `create_env_files` function in `create-env.sh` may not create any configs. The single responsibility of this script is create env files, so let's do it elsewhere

artemiy312: Also `.env` file will be contain the content of `../shopelectro/settings/local.py.dist` (see a 71 line) and it is absolutely unexpected behavior

shopelectro/settings/prod.py
```diff
@@ -1,74 +0,0 @@
...
-REDIS_PASSWORD = os.environ['REDIS_PASSWORD']
-
-REDIS_DSN = f'redis://{REDIS_URL}:{REDIS_PORT}/'
-
-REDIS_LOCATION_DEFAULT = os.environ['REDIS_LOCATION_DEFAULT']
-REDIS_LOCATION_SESSION = os.environ['REDIS_LOCATION_SESSION']
-REDIS_LOCATION_THUMBNAIL = os.environ['REDIS_LOCATION_THUMBNAIL']
-REDIS_LOCATION_USER_AGENT = os.environ['REDIS_LOCATION_USER_AGENT']
-
-CACHES = {
```
artemiy312: I did not understand the point. Where will we store these configs?

docker/create-env.sh
```diff
@@ -51,16 +57,22 @@ Are you ready to continue and build a new .env? [y/n]: " yn
         fi
     fi
 
-    local env_doc=\
+    if [[ ! -f ".env" ]]
```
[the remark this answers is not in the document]
duker33: hmm, it's true we should always overrite it, sorry

docker/create-env.sh
```diff
@@ -51,16 +57,22 @@ Are you ready to continue and build a new .env? [y/n]: " yn
...
+    then
+        echo "generate .env file"
```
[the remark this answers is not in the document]
duker33: we'll remove if, so, we'll remove all nested echo's too

TODO

docker/create-env.sh
duker33: > Also .env file will be contain

y, i'll fix it

> The create_env_files function in create-env.sh

both django settings and env files are about env customisation in terms of our app.
In terms of linux env - not, of course. But app-sided script should work with app env, not with linux env.

I can rename script to `create-local-env`. Or can double it's code to separated script, if it looks confusing

shopelectro/settings/prod.py
duker33: somewhere out of cvs. Simple - on serv and every of us in personal vault. Complex - separated vault or separated cvs repository with configs.

We should remove it by two reasons:
- we should be able to change local (serv too) config without working with cvs
- we should be able to put some secrets to this config

docker/create-env.sh
duker33: moved config processing to separated script

shopelectro/settings/prod.py
artemiy312: > Simple or Complex

We should decide which one to choose. Let's create todo for this

duker33: created https://github.com/fidals/shopelectro/issues/332

docker/create-env.sh
```diff
@@ -24,23 +24,28 @@ function join_by { local IFS="$1"; shift; echo "$*"; }
 
 function create_env_files {
     new_files=()
-    for file in env_files/*.dist
-    do
+    function new_file {
```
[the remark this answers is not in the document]
duker33: TODO

PR #333, opened 2018-06-06
stb181  Doc refarm-site/ source code

duker33:
https://github.com/fidals/stroyprombeton/issues/181


duker33:
Said just few words about refarm-site/ code installation 

See here:
https://github.com/fidals/stroyprombeton/pull/194#issuecomment-394966227

PR #336, opened 2018-06-07
Resurrect coala

artemiy312:


duker33:
@artemiy312 ,

> Work in progress.

`WIP` is standard abbreviation for it. But it's about taste of course.

duker33:
@artemiy312 , it seems we should exclude `node_modules/` folder

```
[WARNING][19:01:47] /node_modules/eslint-config-airbnb-base/rules/imports.js:
	Configuration for rule "import/no-unresolved" is invalid:
	Value "data["0"].caseSensitive" has additional properties.
```

PR #362, opened 2018-06-16
#361  Fix `npm install gulp-imagemin` problem

### LINKED ISSUE #361: Node container build fails
```
gulp build
[14:01:35] gulp-imagemin: Couldn't load default plugin "gifsicle"
```

duker33:
Closes #361

I'll repeat it on STB in https://github.com/fidals/stroyprombeton/issues/203

### DIFF
--- a/docker/images/node/Dockerfile
+++ b/docker/images/node/Dockerfile
@@ -20,5 +20,8 @@ COPY gulpfile.babel.js gulpfile.babel.js
 
 # we use `--no-optional` because some optional npm dependencies fail on install
 RUN npm install -g gulp-cli
+# `npm install gulp-imagemin` don't install it's dependencies sometimes.
+# Took decision from SO: https://goo.gl/3uXZxw
+RUN npm install imagemin-jpegtran imagemin-svgo imagemin-gifsicle imagemin-optipng --save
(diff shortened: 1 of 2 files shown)


docker/images/node/Dockerfile
```diff
@@ -20,5 +20,8 @@ COPY gulpfile.babel.js gulpfile.babel.js
 
 # we use `--no-optional` because some optional npm dependencies fail on install
 RUN npm install -g gulp-cli
+# `npm install gulp-imagemin` don't install it's dependencies sometimes.
+# Took decision from SO: https://goo.gl/3uXZxw
+RUN npm install imagemin-jpegtran imagemin-svgo imagemin-gifsicle imagemin-optipng --save
```
artemiy312: Let's move these dependencies to the `package.json`

duker33: @artemiy312 , i don't want to create mess in `package.json`.
Now it's the good place to look on dependencies and it's versions
But the huge group of `imagemin-*` says nothing useful and takes 50% of `package.json` file.
Nothing useful, because it should be installed recursively from `gulp-imagemin`.

So, it's good to put it to one-lined npm install with it's own clear comment

Issue #331, opened 2018-06-05
Try to use no debug selenium image. stb2


duker33:
Maybe no debug image will decrease time for tests.
Now we have strange output at the end of `selenium` section for every CI. Example: https://ci.fidals.com/fidals/shopelectro/79

- now using: `selenium/standalone-chrome-debug:3.10.0`
- try this: `selenium/standalone-chrome:3.10.0`

duker33:
@liza-andreeva done it

PR #363, opened 2018-06-17
Не используем debug образ selenium

liza-andreeva:
Closes https://github.com/fidals/shopelectro/issues/331


duker33:
we'll fix fell lint step at #366

PR #369, opened 2018-06-19
rf#112  Add .editorconfig

duker33:
#112


.editorconfig
```diff
@@ -0,0 +1,22 @@
+# editorconfig.org
```
artemiy312: Let's add `indent_size = 2` for js/es6 files

PR #381, opened 2018-06-22
#360  Test for wrong redirects behaviour

### LINKED ISSUE #360: Redirects mech broken
Prod db contains these urls in redirects list.
But redirect from first to second is not working

1. https://www.shopelectro.ru/catalog/categories/akkumuliatory-270/tags/6f22/
2. https://www.shopelectro.ru/catalog/categories/tiporazmer-6f22/

duker33:
Closes #360


shopelectro/tests/tests_views.py
```diff
@@ -516,3 +519,32 @@ def test_admin_autocomplete_has_no_model_pages(self):
...
+    fixtures = ['dump.json']
+
+    # will be resurrected at rf#140
+    @unittest.expectedFailure
+    def test_redirect_on_existing_page(self):
+        """DB based redirect from existing url should do, but should not avoid it."""
+        # take some existing `url_from`
+        url_from = '/catalog/categories/category-0/tags/6-v/'
+        response = self.client.get(url_from)
+        self.assertEqual(response.status_code, 200)
```
artemiy312: It is an unnecessary assertion, we should test only redirect logic here

shopelectro/tests/tests_views.py
```diff
@@ -516,3 +519,32 @@ def test_admin_autocomplete_has_no_model_pages(self):
...
+
+
+class Redirects(TestCase):
+
+    fixtures = ['dump.json']
+
+    # will be resurrected at rf#140
+    @unittest.expectedFailure
+    def test_redirect_on_existing_page(self):
+        """DB based redirect from existing url should do, but should not avoid it."""
```
artemiy312: a test is too big and complicated. Let's divide it into smaller pieces

shopelectro/tests/tests_views.py
duker33: @artemiy312 , it's specific of this test case.

`django.contrib.redirects` provide different behaviour if `url_from` returns 200 or 404

shopelectro/tests/tests_views.py
duker33: @artemiy312 , it's single test case.

Do test this case, we should check two things:
`url_from returns 200` and `url_to returns 301`

shopelectro/tests/tests_views.py
```diff
@@ -516,3 +519,32 @@ def test_admin_autocomplete_has_no_model_pages(self):
...
+        # take some existing `url_from`
+        url_from = '/catalog/categories/category-0/tags/6-v/'
```
artemiy312: Let's get a category's url from Model: `Category.objects.first().url`

shopelectro/tests/tests_views.py
artemiy312: @duker33 We can test it separate. 
```python
self.redirect = Redirect.objects.create(  # also we may move it to fixtures
            site=Site.objects.first(),
            old_path=url_from,
            new_path=url_to
        )

def test_redirect_301(self):
    self.assertEqual(self.client.get(self.redirect.old_path).status_code, 300)

def test_following_redirect_200(self):
    self.assertEqual(self.client.get(self.redirect.old_path, follow=True).status_code, 200)
```

shopelectro/tests/tests_views.py
artemiy312: @duker33 but we already test it in other tests, isn't it?

duker33: @artemiy312 , it's new test, that fails.
Maybe we touch redirects in some way in different code. In products scope, for example.
But it's doesn't matter for this task.

Here we test mech for redirects concept itself

shopelectro/tests/tests_views.py
duker33: @artemiy312 , `url_from` should not return 200 in all cases. Only for our single case.
So, this two methods have strong semantical binding. They both mean nothing in separate.

For a different test case, we should create a different test class.

Big test method's size doesn't look bad for me. It consists of simple and sequential steps

shopelectro/tests/tests_views.py
duker33: @artemiy312 , here i should take not the first category. But the one, that has at least one tag.

And it'll be too complex to take it. We have fixtures in cvs, so we can completely rely on them in code

duker33:
discussed directly

requirements.txt
```diff
@@ -1,21 +1,21 @@
...
+selenium==3.11.0
+selenium-requests==1.3
+tblib==1.3.2
+Unidecode==1.0.22
+ua-parser==0.8.0
+user-agents==1.1.0
 sorl-thumbnail==12.4a1
 https://github.com/selwin/django-user_agents/archive/master.zip
-https://github.com/fidals/refarm-site/archive/0.1.6.zip
+https://github.com/fidals/refarm-site/archive/0.1.3.zip
```
artemiy312: It is a wrong version number

shopelectro/tests/tests_views.py
```diff
@@ -522,3 +525,29 @@ def test_admin_autocomplete_has_no_model_pages(self):
...
         self.assertTrue(len(json_to_dict(response)) == 1)
+
+
+class Redirects(TestCase):
+
+    fixtures = ['dump.json']
+
+    # will be resurrected at rf#140
+    @unittest.expectedFailure
+    def test_redirect_on_existing_page(self):
```
artemiy312: Maybe `test_redirect_from_existing_page` will be more relevant. decide by yourself

shopelectro/tests/tests_views.py
```diff
@@ -522,3 +525,29 @@ def test_admin_autocomplete_has_no_model_pages(self):
...
+        """DB based redirect from existing url should do, but should not avoid it."""
```
artemiy312: """DB based redirect from existing url should do, but should not avoid it."""
-> 
"""Redirect middleware of refarm-site redirects from existing url too"""

decide by yourself

requirements.txt
duker33: `0.1.6` fails, as i remember. I'll check it TODO

shopelectro/tests/tests_views.py
duker33: `Refarm-site's custom redirects app should redirect from existing url too`

TODO

shopelectro/tests/tests_views.py
duker33: discussed directly

requirements.txt
```diff
@@ -1,21 +1,21 @@
...
-user-agents==1.0.1
+selenium==3.11.0
+selenium-requests==1.3
+tblib==1.3.2
+Unidecode==1.0.22
+ua-parser==0.8.0
+user-agents==1.1.0
 sorl-thumbnail==12.4a1
 https://github.com/selwin/django-user_agents/archive/master.zip
 https://github.com/fidals/refarm-site/archive/0.1.6.zip
```
artemiy312: You have to bump the refarm-site version to [0.1.7](https://github.com/fidals/refarm-site/releases/tag/0.1.7), isn't it?

duker33: @artemiy312 , nope, 0.1.7 was created after this code

PR #384, opened 2018-06-22
rf#140  Change redirects dep

duker33:
https://github.com/fidals/refarm-site/issues/140

Code bases on branch `360_...` and depends on refarm-site feature. So, this PR merging is blocked but https://github.com/fidals/refarm-site/pull/141, by is ready for review


duker33:
i'll create separated pr with separated task

PR #390, opened 2018-06-26
#389  Save media

### LINKED ISSUE #389: Save image files from test_db. stb2
Now files no saving leads us to 404 errors in tests.
See #359 to find how errors look like

duker33:
Closes #389 

stb2 after this PR


shopelectro/management/commands/test_db.py
```diff
@@ -9,13 +9,13 @@
 - now you have json file, that'll be used by our TDD tests
 """
 import os
-
+import shutil
 from django.conf import settings
```
artemiy312: Please, add a line before third-party packages imports

shopelectro/management/commands/test_db.py
```diff
@@ -118,11 +121,20 @@ def get_name(number, parent=None):
...
             def create_image(file_path, slug):
-                Image.objects.create(
+                # product "/catalog/products/2/" contains image
+                image = Image.objects.create(
                     model=page,
                     slug=slug,
                     image=ImageFile(open(file_path, mode='rb'))
                 )
+                # save files to media folder
+                file_src = open(file_path, mode='rb')
```
artemiy312: Open a file with `with-statement` to release it after use
```python
with open(file_path, mode='rb') as file:
    image = Image.objects.create(
        model=page,
        slug=slug,
        image=ImageFile(file)
    )
    ...
```

shopelectro/management/commands/test_db.py
```diff
@@ -118,11 +121,20 @@ def get_name(number, parent=None):
...
+                file_name = os.path.split(file_src.name)[-1]
```
artemiy312: Let's use [os.path.basepath](https://docs.python.org/3/library/os.path.html#os.path.basename) instead

shopelectro/management/commands/test_db.py
duker33: TODO

shopelectro/management/commands/test_db.py
duker33: TODO

shopelectro/management/commands/test_db.py
duker33: TODO

shopelectro/management/commands/test_db.py
artemiy312: @duker33 the comment above is still not resolved

shopelectro/management/commands/test_db.py
```diff
@@ -118,11 +121,20 @@ def get_name(number, parent=None):
     def create_products(self, categories, tags):
         def create_images(page: Page):
             def create_image(file_path, slug):
-                Image.objects.create(
+                # product "/catalog/products/2/" contains image
+                image = Image.objects.create(
                     model=page,
                     slug=slug,
                     image=ImageFile(open(file_path, mode='rb'))
```
artemiy312: We are opening here the same file as below and does not release it. We should group these lines by one file's context:
```python
with open(file_path, mode='rb') as file_:
    image = Image.objects.create(
        ...,
        image=ImageFile(file_)
    )
    ...
    shutil.copyfile(file_.name, file_dst_path)
```

duker33: Done.

duker33:
created separated task https://github.com/fidals/stroyprombeton/issues/237 for stb

PR #395, opened 2018-07-03
#340 Create pagination buttons and canonical links.

### LINKED ISSUE #340: category_products.html:5: Implement pagination buttons...
The puzzle `302-bdb9bbef` from #302 has to be resolved:

https://github.com/fidals/shopelectro/blob/96e14747c3d7da9dd7db50f01bbb987147e4e2cb/templates/catalog/category_products.html#L5-L5

The puzzle was created by Artemiy on 08-Jun-18. 

Estimate: 60 minutes,     

artemiy312:
The PR does not implement some requirements(2 pagination buttons instead of 5), because we are going to hide pagination buttons from users, @duker33 is it okay?

Closes #340

### DIFF
--- a/shopelectro/tests/tests_views.py
+++ b/shopelectro/tests/tests_views.py
@@ -235,7 +235,7 @@ class CatalogPagination(BaseCatalogTestCase):
 
     @override_settings(DEBUG=True, INTERNAL_IPS=[])
     def test_pagination_buttons(self):
-        """Buttons forward to a previous and a next pagination pages."""
+        """Each button forward to a previous and a next pagination pages."""
         page_number = 3
         prev, next_ = BeautifulSoup(
             self.get_category_page(query_string={'page': page_number}).content.decode('utf-8'),
@@ -245,7 +245,7 @@ class CatalogPagination(BaseCatalogTestCase):
         self.assert_pagination_links(next_, prev, page_number)
 
     def test_pagination_canonical(self):
-        """Links forward to a previous and a next pagination pages."""
+        """Canonical links forward to a previous and a next pagination pages."""
         page_number = 3
         soup = BeautifulSoup(
             self.get_category_page(query_string={'page': page_number}).content.decode('utf-8'),
--- a/shopelectro/tests/tests_views.py
+++ b/shopelectro/tests/tests_views.py
@@ -15,7 +15,7 @@ from bs4 import BeautifulSoup
 from django.conf import settings
 from django.db.models import Q
 from django.http import HttpResponse
-from django.test import TestCase
+from django.test import override_settings, TestCase
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 
@@ -223,9 +223,9 @@ class CatalogPagination(BaseCatalogTestCase):
             404,
         )
 
-    def assert_pagination_links(self, next, prev, page_number):
+    def assert_pagination_links(self, next_, prev, page_number):
         self.assertEqual(
-            get_page_number(self.client.get(next['href'])),
+            get_page_number(self.client.get(next_['href'])),
             page_number + 1,
         )
         self.assertEqual(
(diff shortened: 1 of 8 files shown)


shopelectro/tests/tests_views.py
```diff
@@ -216,6 +216,47 @@ def test_pagination_step(self):
...
         self.assertEqual(len(response.context['product_image_pairs']), pagination_step)
 
+    def test_pagination_404(self):
+        """Category page returns 404 for a nonexistent page number."""
+        self.assertEqual(
+            self.get_category_page(query_string={'page': 1000}).status_code,
+            404,
+        )
+
+    def assert_pagination_links(self, next, prev, page_number):
```
duker33: `next` is reserved word

shopelectro/tests/tests_views.py
```diff
@@ -216,6 +216,47 @@ def test_pagination_step(self):
...
+
+    def test_pagination_buttons(self):
+        """Buttons forward to a previous and a next pagination pages."""
+        page_number = 3
+        prev, next = BeautifulSoup(
+            self.get_category_page(query_string={'page': page_number}).content.decode('utf-8'),
+            'html.parser'
+        ).find(class_='catalog-pagination').find_all('a')
+
+        self.assert_pagination_links(next, prev, page_number)
```
duker33: `self.assert_pagination_links(next, prev, page_number=3)` might be more good, then separated `page_number=3` var with it's state. Decide it

shopelectro/tests/tests_views.py
```diff
@@ -216,6 +216,47 @@ def test_pagination_step(self):
...
+        """Links forward to a previous and a next pagination pages."""
+        page_number = 3
+        soup = BeautifulSoup(
+            self.get_category_page(query_string={'page': page_number}).content.decode('utf-8'),
+            'html.parser'
+        )
+
+        self.assert_pagination_links(
+            soup.find('link', attrs={'rel': 'next'}),
+            soup.find('link', attrs={'rel': 'prev'}),
```
duker33: it'll be good to use explicit args 
```
prev=soup.find(...)
next=soup.find(...)
```

shopelectro/views/catalog.py
```diff
@@ -25,6 +25,13 @@ def get_products_count(request):
     return PRODUCTS_ON_PAGE_MOB if mobile_view else PRODUCTS_ON_PAGE_PC
 
 
+def get_paginated_page(objects, per_page, page_number):
+    try:
+        return Paginator(objects, per_page).page(page_number)
+    except:
```
duker33: common except is not good practice

shopelectro/views/catalog.py
```diff
@@ -25,6 +25,13 @@ def get_products_count(request):
     return PRODUCTS_ON_PAGE_MOB if mobile_view else PRODUCTS_ON_PAGE_PC
 
 
+def get_paginated_page(objects, per_page, page_number):
+    try:
+        return Paginator(objects, per_page).page(page_number)
```
duker33: why not to use `Paginator(...).page(...)` directly?
If we are needed to explain class's method with function wrapper, we have problems with class and/or it's method

templates/catalog/category_pagination.html
```diff
@@ -0,0 +1,19 @@
+{% comment %}
+@todo #302:30m Implement JS for pagination buttons.
+ Hide pagination buttons and reniew a canonical url for current page with the help of JS.
+ Currently buttons are hidden by CSS.
```
duker33: It's dangerous to hide pagination with css.
Wrap pagination with `if debug` plz

shopelectro/views/catalog.py
artemiy312: @duker33 y, I forgot to catch here an `InvalidPage` error

shopelectro/views/catalog.py
artemiy312: @duker33 , 
>why not to use Paginator(...).page(...) directly?

because we have to catch an `InvalidPage` error here.

>we have problems with class and/or it's method

It is a django predefined class, so it is not our an issue.

duker33: @artemiy312 , so, maybe to extend this class with `page_or_404` method?
Decide

PR #401, opened 2018-07-05
stb#233  Forward local.py config according to prod

duker33:
#233

### DIFF
--- a/docker/docker-compose-production.yml
+++ b/docker/docker-compose-production.yml
@@ -20,7 +20,7 @@ services:
     ports:
       - $VIRTUAL_HOST_PORT
     volumes:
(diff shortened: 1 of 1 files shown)


docker/docker-compose-production.yml
```diff
@@ -20,6 +20,7 @@ services:
     ports:
       - $VIRTUAL_HOST_PORT
     volumes:
+      - $SRC_DIR_PROD/shopelectro/settings/local.py:$SRC_DIR/shopelectro/settings/local.py
```
artemiy312: We can mount it in this way: `../shopelectro/settings/local.py`

duker33:
fixed, merge it

Issue #424, opened 2018-07-16
Mail worker broken


duker33:
PO says that mail doesn't work on prod.

`dcp logs celery-mail-worker` run on prod returns traceback:

```
...
se-celery-mail-worker    | Traceback (most recent call last):
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/kombu/transport/base.py", line 123, in __getattr__
se-celery-mail-worker    |     return self[key]
se-celery-mail-worker    | KeyError: 'async'
se-celery-mail-worker    |
se-celery-mail-worker    | During handling of the above exception, another exception occurred:
se-celery-mail-worker    |
se-celery-mail-worker    | Traceback (most recent call last):
se-celery-mail-worker    |   File "/usr/local/bin/celery", line 11, in <module>
se-celery-mail-worker    |     sys.exit(main())
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/__main__.py", line 14, in main
se-celery-mail-worker    |     _main()
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/celery.py", line 326, in main
se-celery-mail-worker    |     cmd.execute_from_commandline(argv)
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/celery.py", line 488, in execute_from_commandline
se-celery-mail-worker    |     super(CeleryCommand, self).execute_from_commandline(argv)))
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/base.py", line 281, in execute_from_commandline
se-celery-mail-worker    |     return self.handle_argv(self.prog_name, argv[1:])
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/celery.py", line 480, in handle_argv
se-celery-mail-worker    |     return self.execute(command, argv)
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/celery.py", line 412, in execute
se-celery-mail-worker    |     ).run_from_argv(self.prog_name, argv[1:], command=argv[0])
se-celery-mail-worker    |   File "/usr/local/lib/python3.6/site-packages/celery/bin/worker.py", line 221, in run_from_argv
se-celery-mail-worker    |     return self(*args, **options)
[...]

PR #425, opened 2018-07-16
#424  Update celery

duker33:


artemiy312:
@duker33 Should we repeat the same step for STB?

PR #427, opened 2018-07-18
stb#260  Rm redundant services ports forwarding

duker33:
https://github.com/fidals/stroyprombeton/issues/260


duker33:
merge it

PR #443, opened 2018-07-24
#273 Do not remove products

### LINKED ISSUE #273: Похожие товары на 404 странице
На 404 странице под кнопкой "На главную" выводим  10 рандомных товаров их то же категории, где был удаленный товар. 
С заголовком "Похожие товары"
Две строки по 5 товаров. как на главной http://prntscr.com/ivclw1

artemiy312:
Closes #273 

Prevent products removing during 1C update

### DIFF
--- a/shopelectro/management/commands/excel.py
+++ b/shopelectro/management/commands/excel.py
@@ -112,7 +112,7 @@ class Command(BaseCommand):
         def write_product_rows():
             """Write products lines."""
             sheet = self.sheet
-            products = Product.objects.filter(category=category)
+            products = Product.objects.filter(category=category, page__is_active=True)
             for product in products.iterator():
                 product_start = 'A' + self.CURRENT_ROW
                 sheet[product_start] = product.name
--- a/shopelectro/views/catalog.py
+++ b/shopelectro/views/catalog.py
@@ -57,6 +57,7 @@ class ProductPage(catalog.ProductPage):
             # @todo #273 Render 404, that is recommending products for a deleted product.
             #  1. Find a product with page__is_active=False
             #  2. If the product exists then render 404 with products of its category
+            #  See the parent issue for details.
             raise error404
         context = self.get_context_data(object=self.object)
         return self.render_to_response(context)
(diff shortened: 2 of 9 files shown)


duker33:
@artemiy312 , it's not resolve issue. Leave pdd subtasks for 404 plz

artemiy312:
@duker33 I already have [created it](https://github.com/fidals/shopelectro/pull/443/files#diff-9e53a25fb3c9eb1745ff37a5799ef9daR57)

shopelectro/management/commands/excel.py
```diff
@@ -112,7 +112,7 @@ def collapse_row(row):
         def write_product_rows():
             """Write products lines."""
             sheet = self.sheet
-            products = Product.objects.filter(category=category)
+            products = Product.objects.filter(category=category, page__is_active=True)
```
duker33: let's add custom ModelManager instead:

```
class ProductManager(models.Manager):
    def get_queryset(self):
        return super(DahlBookManager, self).get_queryset().filter(author='Roald Dahl')

class Product:
    active = ProductManager()

# in client code
Product.active.filter(category=category)
```

Took example here: https://docs.djangoproject.com/en/1.11/topics/db/managers/#modifying-a-manager-s-initial-queryset

shopelectro/views/catalog.py
```diff
@@ -45,11 +45,22 @@ class ProductPage(catalog.ProductPage):
...
         .select_related('page')
     )
 
+    def get(self, request, *args, **kwargs):
+        try:
+            self.object = self.get_object()
+        except http.Http404 as error404:
+            # @todo #273 Render 404, that is recommending products for a deleted product.
+            #  1. Find a product with page__is_active=False
+            #  2. If the product exists then render 404 with products of its category
```
duker33: Let's add this:
```
...
3. See parent task for details.
```

Some details about design are missed in this task body

shopelectro/management/commands/excel.py
artemiy312: @duker33 I am going to create todo for it and it seems that this code will not work, but we may discuss it in the related issue

PR #452, opened 2018-07-27
#fix  Return 1C tests

duker33:
Signed-off-by: duker33 <duker33@gmail.com>


0pdd:
@duker33 the puzzle [#460](https://github.com/fidals/shopelectro/issues/460) is still not solved.

PR #499, opened 2018-08-09
#388  Product page siblings

### LINKED ISSUE #388: Products re-linking as on stb
We already have this functinality on stb.

[Reqs from seo guys](https://docs.google.com/document/d/14ZcX5ujkQdIHfZ1tL31Nr1cHnwJ5N037Blsqf5A0jpE/edit?ts=5b13c914#).

Screen [is from this page](http://www.stroyprombeton.ru/gbi/products/638/)

image

duker33:
Closes #388

### DIFF
--- a/shopelectro/views/catalog.py
+++ b/shopelectro/views/catalog.py
@@ -152,7 +152,7 @@ class IndexPage(pages_views.CustomPageView):
                 .prefetch_related('category')
                 .select_related('page')
             )
-            tile_products = prepare_tile_products(top_products)
+            tile_products = ProductsAdapter(top_products).prepare()
 
         return {
             **context,
--- a/shopelectro/views/catalog.py
+++ b/shopelectro/views/catalog.py
@@ -39,15 +39,19 @@ class CategoryTree(catalog.CategoryTree):
     category_model = models.Category
 
 
-class ProductsAdapter:
-    """Adapts products QuerySet for using in template."""
+class TemplateDataAdapter:
+    """
+    Adapts data to template format.
+
+    In most cases data came from DB.
+    """
 
     def __init__(self, products: QuerySet):
         self.products = products
 
-    # @todo #388:15m Annotate method `ProductsAdapter.prepare`.
+    # @todo #388:15m Annotate method `TemplateDataAdapter.adapt_products`.
     #  With namedtuple, dataclass or smth.
-    def prepare(self):
+    def adapt_products(self):
         images = Image.objects.get_main_images_by_pages(
             models.ProductPage.objects.filter(
                 shopelectro_product__in=self.products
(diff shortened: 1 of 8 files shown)


shopelectro/views/catalog.py
```diff
@@ -38,19 +39,27 @@ class CategoryTree(catalog.CategoryTree):
     category_model = models.Category
 
 
-def prepare_tile_products(products):
-    images = Image.objects.get_main_images_by_pages(
-        models.ProductPage.objects.filter(
-            shopelectro_product__in=products
+class ProductsAdapter:
```
artemiy312: What does the `Adapter` mean here? It seems like nothing, because we may say, that every piece of code adapts something for something else. So let's call it more verbose like `TileProducts` (Tile contains product's name, image and category). Decide yourself

shopelectro/views/catalog.py
```diff
@@ -38,19 +39,27 @@ class CategoryTree(catalog.CategoryTree):
...
+    """Adapts products QuerySet for using in template."""
+
+    def __init__(self, products: QuerySet):
+        self.products = products
+
+    # @todo #388:15m Annotate method `ProductsAdapter.prepare`.
+    #  With namedtuple, dataclass or smth.
+    def prepare(self):
```
artemiy312: `TileProducts(...).get()` also would sound better

shopelectro/views/catalog.py
duker33: @artemiy312 , it's not about tile products, but about converting one QuerySet to another for template format. We can move every queryset from such classes instead of using huge `prepare` inner functions in views.
I'll suggest different approach here

shopelectro/views/catalog.py
duker33: @artemiy312 , it's confusing to see QuerySet as output of get method.
I'll suggest some differ approach

artemiy312: @duker33 Result of the method is list of tuples

shopelectro/views/catalog.py
```diff
@@ -38,19 +39,31 @@ class CategoryTree(catalog.CategoryTree):
     category_model = models.Category
 
 
-def prepare_tile_products(products):
-    images = Image.objects.get_main_images_by_pages(
-        models.ProductPage.objects.filter(
-            shopelectro_product__in=products
+class TemplateDataAdapter:
```
artemiy312: @duker33 
I still believe that such names are very bad for any classes, because they do not make any sense for us. I still would prefer to see `ProductCards/ProductTiles(...).get()` or something like that.

Decide

Also Yegor B. has an [article](https://www.yegor256.com/2015/03/09/objects-end-with-er.html) about such names :)

duker33: @artemiy312 , my idea is not about products cards or tiles.
The idea is to systemize bunch of `prepare` methods with classes hierarchy.
But seems i don't know how to it right now.
I'll revert code related to this arch refactoring and create a separated issue about it

PR #501, opened 2018-08-10
stb#263 Add secrets: option to .drone.yml

artemiy312:
Closes fidals/stroyprombeton#263

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -45,6 +45,7 @@ pipeline:
 
   test:
     image: fidals/se:dev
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -151,14 +150,12 @@ services:
       event: [push, pull_request]
 
   selenium:
-    image: selenium/standalone-chrome-debug:3.10.0
+    image: selenium/standalone-chrome:3.10.0
```
duker33: We should use debug here, see related comment
https://github.com/fidals/stroyprombeton/pull/291#discussion_r209157758

PR #509, opened 2018-08-13
#265  Remove config

### LINKED ISSUE #265: Remove config.py
Connected with fidals/refarm-site#89

Снесём config.py. Переносим его содержимое в settings, utils и т.д.

MAIN_PAGE_TILE делаем отдельной сущностью в базе и прикручиваем к админке.

duker33:
Closes #265


shopelectro/views/catalog.py
```diff
@@ -33,6 +32,18 @@ def get_paginated_page_or_404(objects, per_page, page_number):
         raise http.Http404('Page does not exist')
 
 
+class CategorySortingOption:
+    def __init__(self, index=0):
```
artemiy312: >index

seems `number` may be more relevant, because "option number" sounds better then "option index". Decide yourself

shopelectro/views/catalog.py
```diff
@@ -33,6 +32,18 @@ def get_paginated_page_or_404(objects, per_page, page_number):
...
+        options = settings.CATEGORY_SORTING_OPTIONS[index]
+        self.label = options['label']
+        self.field = options['field']
+        self.direction = options['direction']
+
+    @property
+    def directed_field(self):
```
artemiy312: `as_db_exp` may be better

shopelectro/views/catalog.py
```diff
@@ -176,7 +187,7 @@ def get_context_data(self, **kwargs):
         page_number = int(self.request.GET.get('page', 1))
         view_type = self.request.session.get('view_type', 'tile')
         sorting = int(self.kwargs.get('sorting', 0))
-        sorting_option = config.category_sorting(sorting)
+        sorting_option_str = CategorySortingOption(index=sorting).directed_field
```
artemiy312: `sorting_option_str` -> `sorting_option: str`

shopelectro/views/catalog.py
```diff
@@ -213,7 +224,7 @@ def get_context_data(self, **kwargs):
                 .filter(tags__in=tags)
                 # Use distinct because filtering by QuerySet tags,
                 # that related with products by many-to-many relation.
-                .distinct(sorting_option.lstrip('-'))
+                .distinct(sorting_option_str.lstrip('-'))
```
artemiy312: `sorting_option_str.lstrip('-')` -> 
```python
option = CategorySortingOption(...)
...
.distinct(option.field_name)

File: shopelectro/views/catalog.py [hunk 5]
```diff
@@ -33,6 +32,18 @@ def get_paginated_page_or_404(objects, per_page, page_number):
         raise http.Http404('Page does not exist')
 
 
+class CategorySortingOption:
```
artemiy312: `SortingOption` may be sufficient, because we use it only in `catalog.py`

PS: by the way it seems it is about a catalog sorting and not about a category

shopelectro/views/catalog.py
```diff
@@ -242,7 +253,7 @@ def template_context(page, tag_titles, tags):
             'total_products': total_products,
             'products_count': (page_number - 1) * products_on_page + products.count(),
             'paginated_page': paginated_page,
-            'sorting_options': config.category_sorting(),
+            'sorting_options': settings.CATEGORY_SORTING_OPTIONS.values(),
```
artemiy312: It may be better to use it as a method of `CategorySortingOption`

shopelectro/views/catalog.py
duker33: @artemiy312 , y, i thought about smth like this. But this approach has several cons:
- it's not db (sql) exp, but query_set's one
- `exp` or `qs` are shortlands, it'll be good to name smth without it
- `name` or `-name` is too simple concept to explain it in query set terms

I'll add pydoc with examples instead

shopelectro/views/catalog.py
duker33: @artemiy312 , y, it'll be good to hide every sorting options processing logic in `SortingOption` abstraction.
But `SortingOption` can't do this himself in this case.

We should create class `SortingOptions` for it. And right now it'll have only one purpose: call dict `values` method inside. And it'll have only one use case here.
So, here i use settings var directly

shopelectro/views/catalog.py
duker33: thought about it. Indexing - that's pretty what we are doing with sorting options.
Number is some semantical thing. But price doesn't have index/number value `0` buy it's semantic

PR #521, opened 2018-08-16
#481  Create new method

### LINKED ISSUE #481: catalog.py:189-190: Implement a new method for a...
The puzzle `470-c5d3123c` from #470 has to be resolved:

https://github.com/fidals/shopelectro/blob/814eee587fcffe14d3c2ca276eb50f53fb0af0a4/shopelectro/views/catalog.py#L189-L190

The puzzle was created by Artemiy on 06-Aug-18. 

Estimate: 15 minutes,     

duker33:
Closes #481

### DIFF
--- a/shopelectro/sitemaps.py
+++ b/shopelectro/sitemaps.py
@@ -49,7 +49,7 @@ def get_categories_with_tags() -> Generator[
     So, for each tags group in each category we'll get 1 tag.
     """
     for category in Category.objects.filter(page__is_active=True):
-        products = Product.objects.get_by_category(category)
+        products = Product.objects.get_category_descendants(category)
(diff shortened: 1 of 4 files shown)


shopelectro/sitemaps.py
```diff
@@ -49,7 +49,7 @@ def get_categories_with_tags() -> Generator[
     So, for each tags group in each category we'll get 1 tag.
     """
     for category in Category.objects.filter(page__is_active=True):
-        products = Product.objects.get_by_category(category)
+        products = Product.objects.get_category_descendants(category)
```
artemiy312: Seems here will be enough `get_by_category` and below too

PR #527, opened 2018-08-17
#522  Make tag slugs autounique

### LINKED ISSUE #522: tests_views.py:190-193: Fix doubled tags issue....
The puzzle `398-023fcade` from #398 has to be resolved:

https://github.com/fidals/shopelectro/blob/6806692e83bba09f669baf022ed379e56d7e0b4d/shopelectro/tests/tests_views.py#L190-L193

The puzzle was created by duker33 on 17-Aug-18. 

Estimate: 60 minutes,     

duker33:
Closes #522

### DIFF
--- a/shopelectro/models.py
+++ b/shopelectro/models.py
@@ -24,6 +24,13 @@ from ecommerce.models import Order as ecOrder
 from pages.models import CustomPage, ModelPage, Page, SyncPageMixin, PageManager
 
 
+def randomize_slug(slug: str) -> str:
+    slug_hash = ''.join(
+        random.choices(string.ascii_lowercase, k=settings.SLUG_HASH_SIZE)
+    )
+    return f'{slug}_{slug_hash}'
+
+
 class SECategoryQuerySet(TreeQuerySet):
     def get_categories_tree_with_pictures(self) -> 'SECategoryQuerySet':
         categories_with_pictures = (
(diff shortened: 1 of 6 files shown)


shopelectro/tests/tests_commands.py
```diff
@@ -130,6 +139,23 @@ def test_update_or_create_tags(self):
...
+    #  Resurrect test `shopelectro.tests.tests_views.CatalogTags#test_doubled_tag`
+    @unittest.expectedFailure
+    def test_create_double_named_tags(self):
+        """Two tags with the same name should have different slugs."""
+        # create two tags with the same name, but from different groups
+        tag_data = {
+            **get_tag_as_dict(group='First group', tag='Doubled tag'),
+            **get_tag_as_dict(group='Second group', tag='Doubled tag'),
+        }
+        update_tags.create_or_update(tag_data)
```
artemiy312: Why do you associate this test with catalog updating command? I believe, that it will better to add unique constraint to `name` field of `Tag` model and move this test to `tests_models.py`

duker33: @artemiy312 , y, that's what i've done implementing this subtask.
I'll propose the whole decision in this PR

duker33:
implemented problem solving

shopelectro/models.py
```diff
@@ -232,6 +236,20 @@ def get_group_tags_pairs(self) -> List[Tuple[TagGroup, List['Tag']]]:
 
         return group_tags_pair
 
+    def create_safely(self, name: str, slug='', **kwargs):
```
artemiy312: @duker33 
Maybe should we do it in `Tag.objects.save(...)` method to provide safely creation for every cases?

duker33: hmm, i'll think about it during next day

PR #531, opened 2018-08-19
#525 jsecommerce brand

### LINKED ISSUE #525: tracking.es6:35: Send info about product's brand to YA...
The puzzle `504-f7f52310` from #504 has to be resolved:

https://github.com/fidals/shopelectro/blob/4e607fb856842b43192511b1b4eaff19d2052b9a/front/js/shared/tracking.es6#L35-L35

The puzzle was created by Artemiy on 17-Aug-18. 

Estimate: 30 minutes,     

artemiy312:
Closes #525


duker33:
see no `WIP` tag here, so i assume, that i should do review

shopelectro/tests/tests_models.py
```diff
@@ -21,3 +24,42 @@ def test_creation_deactivated_product(self):
...
+            self.assertIn(product, tag.products.all())
+
+    def test_filter_by_products_equality(self):
+        products = self.get_products()
+        self.assertEquals(
+            set(Tag.objects.filter_by_products(products)),
+            set(chain.from_iterable(product.tags.all() for product in products)),
+        )
+
+    def test_filter_by_products_quantity(self):
```
duker33: equality and quantity both are props of the same structure.
It'll be more correct to test it with one case, i think.

Draft code proposal:
```
sorted_by_id = partial(sorted(key=lambda x: x.id))

self.assertEquals(
            sorted_by_id(list(Tag.objects.filter_by_products(products))),
            sorted_by_id(list(chain.from_iterable(product.tags.all() for product in products))),
        )
```

Decide yourself

shopelectro/tests/tests_models.py
```diff
@@ -21,3 +24,42 @@ def test_creation_deactivated_product(self):
...
+            len(set(chain.from_iterable(product.tags.all() for product in products))),
+        )
+
+
+class QueriesNumber(TransactionTestCase):
+
+    fixtures = ['dump.json']
+
+    def test_tag_get_brands(self):
+        """Perform two query to fetch tags and related products."""
```
duker33: `two queries`

shopelectro/tests/tests_models.py
```diff
@@ -21,3 +24,42 @@ def test_creation_deactivated_product(self):
...
+
+    def test_filter_by_products_quantity(self):
+        products = self.get_products()
+        self.assertEquals(
+            Tag.objects.filter_by_products(products).count(),
+            len(set(chain.from_iterable(product.tags.all() for product in products))),
+        )
+
+
+class QueriesNumber(TransactionTestCase):
```
duker33: this class has the semantical intersection with `TagModel`.
It's not clear where we should put `test_tag_save` method: to this class or to `TagModel`.

I propose to call it `TagQuerySet`

templates/catalog/category_products.html
```diff
@@ -43,11 +43,12 @@
         {% endcomment %}
         <button class="btn btn-blue btn-category-buy js-product-to-cart"
                 productId="{{ product.id }}" productPrice="{{ product.price }}"
-                productName="{{ product.name }}">
+                productName="{{ product.name }}"
+                productBrand="{% if brand %}{{ brand.name }}{% else %}{% endif %}">
```
duker33: `{% else %}` is redundant

shopelectro/models.py
```diff
@@ -121,6 +120,10 @@ def get_siblings(self, offset):
             .select_related('page')[:offset]
         )
 
+    def get_brand_name(self) -> str:
```
duker33: if we decide to create `get_brand` method, we can rm it.
See get_brand method proposal [here](https://github.com/fidals/shopelectro/pull/531/files#diff-363c06261e0934f61f73b3e7004f01c4R237)

shopelectro/tests/tests_models.py
artemiy312: I am going to create docstring for uit

shopelectro/tests/tests_models.py
artemiy312: @duker33 We have to use `set(...)` in this case, because  `chain.from_iterable(product.tags.all() for product in products)` should contain unique tags, therefore we can not use `sorted_by_id(list(..))`

shopelectro/models.py
```diff
@@ -213,24 +216,36 @@ def __str__(self):
...
+    def get_group_tags_pairs(self) -> List[Tuple[TagGroup, List['Tag']]]:
+        grouped_tags = groupby(self.prefetch_related('group'), key=attrgetter('group'))
+        return [
             (group, list(tags_))
-            for group, tags_ in groupby(tags, key=attrgetter('group'))
+            for group, tags_ in grouped_tags
         ]
 
-        return group_tags_pair
+    def get_brands(self, products: List[Product]) -> Dict[Product, 'Tag']:
```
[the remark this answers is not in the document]
artemiy312: It is required a lot more queries then `Tag.get_brands`, because of django ORM :)
I am going to create test to show it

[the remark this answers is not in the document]
duker33: @artemiy312 , then just leave pdd task with link on this thread plz

shopelectro/tests/tests_models.py
```diff
@@ -39,27 +40,35 @@ def test_get_brands_content(self):
...
+    def test_get_brands_from_tags(self):
+        """Perform two queries to fetch tags, its products."""
+        product = list(Product.objects.all())
         with self.assertNumQueries(2):
-            Tag.objects.get_brands(products)
+            Tag.objects.get_brands(product)
+
+    def test_get_brands_from_products(self):
+        """Perform a lot of queries, so we should fetch brands from the Tag model."""
+        products = Product.objects.all().prefetch_related('tags', 'tags__group')
```
duker33: emm, but what about transaction.Atomic?

artemiy312: @duker33 How is a transaction related to the number of queries? Transaction is about db consistency only

duker33: @artemiy312 , y, it's true, sorry

PR #532, opened 2018-08-21
#524 jsecommerce purchase

### LINKED ISSUE #524: tracking.es6:30-33: Send `purchase` event to YA and GA...
The puzzle `504-30170dc3` from #504 has to be resolved:

https://github.com/fidals/shopelectro/blob/4e607fb856842b43192511b1b4eaff19d2052b9a/front/js/shared/tracking.es6#L30-L33

The puzzle was created by Artemiy on 17-Aug-18. 

Estimate: 60 minutes,     

artemiy312:
Closes #524

### DIFF
--- a/front/js/components/order.es6
+++ b/front/js/components/order.es6
@@ -48,7 +48,7 @@
       'onCartUpdate', renderTable, fillSavedInputs,
       touchSpinReinit, restoreSelectedPayment, cityAutocomplete,
     );
-    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend', [getProductsData()]));
+    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend'));
 
     /**
      * Bind events to parent's elements, because of dynamic elements.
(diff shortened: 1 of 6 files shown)


front/js/components/order.es6
```diff
@@ -48,7 +48,7 @@
       'onCartUpdate', renderTable, fillSavedInputs,
       touchSpinReinit, restoreSelectedPayment, cityAutocomplete,
     );
-    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend', [getProductsData()]));
+    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend'));
```
duker33: why you preferred to send data with via backend, not frontend mediator?
Mediator approach seems more explicit

shopelectro/views/ecommerce.py
```diff
@@ -47,7 +48,20 @@ class FlushCart(ec_views.FlushCart):
 
 
 class OrderSuccess(ec_views.OrderSuccess):
-    order = Order
+    order = Order.objects.all().prefetch_related('positions')
```
duker33: we should send only one success order, as i see

front/js/components/order.es6
artemiy312: @duker33 We still use mediator to publish onOrderSend event on submit. What did you mean?

shopelectro/views/ecommerce.py
artemiy312: @duker33 We still send only one success order with its positions

front/js/components/order.es6
duker33: >  [getProductsData()]

@artemiy312 , i mean this data. We use it here to trigger event, but not for sending data

artemiy312: @duker33 Now we have to send data on success-page, because we need a order's id for jsecommerce

shopelectro/views/ecommerce.py
duker33: @artemiy312 , y, you are right. It's DetailedView, sorry

Issue #537, opened 2018-08-22
tests_models.py:85-90: Try to implement...


0pdd:
The puzzle `525-5d17d280` from #525 has to be resolved:

https://github.com/fidals/shopelectro/blob/64939d2fe050fd9c4cbd110be802a9a0444ca5aa/shopelectro/tests/tests_models.py#L85-L90

The puzzle was created by Artemiy on 22-Aug-18. 

Estimate: 30 minutes,     

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@artemiyds2 @duker33 please pay attention to this new issue.

artemiy312:
I'd prefer to leave `Tag.objects.get_brands()`, because a brand is `Tag` object

duker33:
@artemiy312 , maybe then Product.objects.get_brands() ?
Props:
- we don't needed method arg `products` in client code
- we don't needed in Tag class in client code: `Tag.objects.get_brands(products.get_brands())` -> `products.get_brands()`

artemiy312:
@duker33 

`products.get_brands()` maybe looks more convenient, but will work ineffective as i showed through [test](https://github.com/fidals/shopelectro/blob/64939d2fe050fd9c4cbd110be802a9a0444ca5aa/shopelectro/tests/tests_models.py#L85-L90)

PR #551, opened 2018-08-27
#550  Split TaggedCategoryPage class

### LINKED ISSUE #548: Return E305 pycodestyle rule
currently coala's config contains this code string:
`pycodestyle_ignore = E305,W503`

But it's pep8's rule, as Guide says here: https://github.com/PyCQA/pycodestyle/issues/400

### LINKED ISSUE #550: Split CatalogPage's monolith code
shopelectro.views.catalog.CategoryPage currently consists of this steps:
- get products
- filter tags
- get template context
- paginate products

All these steps are semantically separated from CategoryPage and from each other.
Here is good code draft example for `get products` + `filter tags`.

```
class CategoryPage:

    def __init__(self, category_id):
        pass

    def products(self) -> 'QuerySet':
        return None

    def get_context_data(self):
        return {}


class TaggedCategoryPage:
    def __init__(self, page: CategoryPage, tags):
        self.super = page

    def products(self) -> 'QuerySet':
        return super.products.filter(smth=anyth)

    def get_context_data(self):
        context = self.super.get_context_data()
        data = {}
        return {
            'all_products': products,
            **data,
            **context,
        }
```

But use in this task traditional inheritance instead of objects composition

duker33:
It's the start of splitting CatalogPage's monolith code

Closes #550 
Closes #548

### DIFF
--- a/shopelectro/context.py
+++ b/shopelectro/context.py
@@ -1,0 +1,368 @@
+"""
+Contains view context classes.
+
+Context is our own concept.
+It's part of the View abstract layer in MTV paradigm.
+Data flow looks like this:
+Urls -> Context -> View.
+So, by concept, View receives context directly
+without accessing url args.
+It stands between urls and view functions.
+
+Every context class is used via objects composition.
+Code example to create tagged category:
+
+>>> from django.http import HttpRequest
+>>> url_kwargs = {'slug': 'my-page'}
+>>> Category(url_kwargs, request=HttpRequest()) | TaggedCategory()
+"""
+
+import typing
+from abc import ABC, abstractmethod
+from functools import lru_cache, partial
+
+from django import http
+from django.db.models import QuerySet
+from django.conf import settings
+from django.core.paginator import Paginator, InvalidPage
+from django_user_agents.utils import get_user_agent
+
+from images.models import Image
+from pages.models import ModelPage
+
+from shopelectro import models
+
+
+class SortingOption:
+    def __init__(self, index=0):
+        options = settings.CATEGORY_SORTING_OPTIONS[index]
+        self.label = options['label']
+        self.field = options['field']
+        self.direction = options['direction']
+
+    @property
+    def directed_field(self):
+        return self.direction + self.field
+
+
+class PaginatorLinks:
+
+    def __init__(self, number, path, paginated: Paginator):
+        self.paginated = paginated
+        self.number = number
+        self.path = path
+
+        self.index = number - 1
+        self.neighbor_bounds = settings.PAGINATION_NEIGHBORS // 2
+        self.neighbor_range = list(self.paginated.page_range)
+
+    def page(self):
+        try:
+            return self.paginated.page(self.number)
+        except InvalidPage:
+            raise http.Http404('Page does not exist')
+
+    def showed_number(self):
+        return self.index * self.paginated.per_page + self.page().object_list.count()
+
+    def _url(self, number):
+        self.paginated.validate_number(number)
+        return self.path if number == 1 else f'{self.path}?page={number}'
+
+    def prev_numbers(self):
+        return self.neighbor_range[:self.index][-self.neighbor_bounds:]
+
+    def next_numbers(self):
+        return self.neighbor_range[self.index + 1:][:self.neighbor_bounds]
+
+    def number_url_map(self):
+        numbers = self.prev_numbers() + self.next_numbers()
+        return {number: self._url(number) for number in numbers}
+
+
+# @todo #550:30m Split to ProductImagesContext and ProductBrandContext
+@lru_cache(maxsize=64)
+def prepare_tile_products(products: QuerySet):
+    assert isinstance(products, QuerySet)
+
+    images = Image.objects.get_main_images_by_pages(
+        models.ProductPage.objects.filter(shopelectro_product__in=products)
+    )
+
+    brands = (
+        models.Tag.objects
+        .filter_by_products(products)
+        .get_brands(products)
+    )
+
+    return [
+        (product, images.get(product.page), brands.get(product))
+        for product in products
+    ]
+
+
+class ObjectsComposition:
+
+    super: 'ObjectsComposition' = None
+
+    def __or__(self, other: 'ObjectsComposition'):
+        other.super = self
+        return other
+
+
+# @todo #550:120m Move context realization to pure to objects composition.
+#  Discussed some thoughts with Artemiy via call.
+#  Artemiy will do it.
+#  For example SortedCategory should
+#  consist of separated SortedList and Category classes/objects.
+class AbstractContext(ObjectsComposition, ABC):
+
+    super: 'AbstractContext' = None
+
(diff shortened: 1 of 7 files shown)


shopelectro/views/catalog.py
```diff
@@ -178,49 +179,75 @@ def merge_products_context(products):
...
-            all_products = (
-                all_products
+# @todo #550:60m Split TaggedCategoryPage class
+#  See details at parent task.
+class TaggedCategoryPage(CategoryPage):
+
+    def get_sorting_index(self):
+        return int(self.kwargs.get('sorting', 0))
+
+    def get_tags(self) -> typing.Union[models.TagQuerySet, None]:
```
artemiy312: typing.Optional[models.TagQuerySet] will be better

shopelectro/views/catalog.py
```diff
@@ -178,49 +179,75 @@ def merge_products_context(products):
...
-            slugs = models.Tag.parse_url_tags(tags)
-            tags = models.Tag.objects.filter(slug__in=slugs)
 
-            all_products = (
-                all_products
+# @todo #550:60m Split TaggedCategoryPage class
+#  See details at parent task.
+class TaggedCategoryPage(CategoryPage):
+
+    def get_sorting_index(self):
```
artemiy312: Seems the better place for it in `CategoryPage` class

shopelectro/views/catalog.py
```diff
@@ -229,26 +256,28 @@ def template_context(page, tag_titles, tags):
                 'tags': tags,
             }
 
-        page = context['page']
-        page.get_template_render_context = partial(
-            template_context, page, tag_titles, tags)
+        tags = self.get_tags()
```
artemiy312: You are calling it two times: here and in `get_products(...)`. It will increase quantity of queries to db.

shopelectro/views/catalog.py
```diff
@@ -229,26 +256,28 @@ def template_context(page, tag_titles, tags):
...
-        total_products = all_products.count()
+        paginated_page = get_paginated_page_or_404(products, products_on_page, page_number)
+        total_products = products.count()
         products = paginated_page.object_list
         if not products:
             raise http.Http404('Page without products does not exist.')
 
         return {
             **context,
             'products_data': merge_products_context(products),
```
artemiy312: It will increase quantity of queries too, because you are already doing it in `CategoryPage` class

shopelectro/views/catalog.py
duker33: @artemiy312 , the best place is SortedCategoryPage.
I moved all instead of CategoryPage's logic to TaggedCategoryPage.
Then i'll fork another class from TaggedCategoryPage and so on

shopelectro/views/catalog.py
duker33: @artemiy312 , i thought, ORM caches such queries. I'll fix it

duker33:
@artemiy312 , review plz again

shopelectro/views/catalog.py
```diff
@@ -178,83 +182,27 @@ def merge_products_context(products):
...
 
-        group_tags_pairs = (
-            models.Tag.objects
-            .filter_by_products(all_products)
-            .get_group_tags_pairs()
+    def get_context_data(self, **kwargs):
+        """Add sorting options and view_types in context."""
+        context_ = (
+            context.Category(self.kwargs, self.request)
+            | context.TaggedCategory()
```
artemiy312: Currently I do not understand why we need such syntax: `context_obj | context_obj | ...`.
It has some critical cons:
- We can not explicitly work with dependencies
- We have to note, how some context depends on another one

I propose to use classic object composition, that has no any of these cons

artemiy312: e.g. of object composition:
```
category = context.SortedCategory(context.TaggedCategory(context.Category(self.kwargs, self.request)))
pagination = context.PaginationContext(category)
rendered_fields = context.RenderedFields(category)

Contexts(category, pagination, rendered_fields).get_context()
```

shopelectro/context.py
```diff
@@ -0,0 +1,372 @@
...
+        self.label = options['label']
+        self.field = options['field']
+        self.direction = options['direction']
+
+    @property
+    def directed_field(self):
+        return self.direction + self.field
+
+
+# @todo #539:60m Move PaginatorLinks to refarm-site.
```
artemiy312: This is no longer relevant todo, i am going to integrate implemented `refarm_pagination` after https://github.com/fidals/refarm-site/issues/175

shopelectro/context.py
```diff
@@ -0,0 +1,372 @@
...
+from abc import ABC, abstractmethod
+from functools import lru_cache, partial
+
+from django import http
+from django.db.models import QuerySet
+from django.conf import settings
+from django.core.paginator import Paginator, InvalidPage
+from django_user_agents.utils import get_user_agent
+
+from images.models import Image
```
artemiy312: We should place RF related packages to single group, as i remember

shopelectro/context.py
```diff
@@ -0,0 +1,375 @@
...
+            self.page.model
+        )
+
+    def get_context_data(self):
+        """Add sorting options and view_types in context."""
+        # @todo #550:15m Take `view_type` value from dataclass.
+        #  Depends on updating to python3.7
+        view_type = self.request.session.get('view_type', 'tile')
+
+        # @todo #550:60m Create `Product.get_brand`
```
artemiy312: We already have [the issue](https://github.com/fidals/shopelectro/issues/537) for it

shopelectro/context.py
```diff
@@ -0,0 +1,375 @@
...
+            'view_type': view_type,
+        }
+
+
+class TaggedCategory(AbstractProductsListContext):
+
+    def get_sorting_index(self):
+        return int(self.url_kwargs.get('sorting', 0))
+
+    # @todo #550:30m Test if context.TaggedCategory.get_tags cached.
```
artemiy312: What exactly  should be cached here?

artemiy312:
@duker33 I left some comments

shopelectro/context.py
duker33: @artemiy312 , y, i supposed them. But django ORM caches them by default.
So, i rm both lru_cache and task

shopelectro/context.py
```diff
@@ -0,0 +1,368 @@
+"""
+Contains view context classes.
+
+Context is our own concept.
+It's part of the View abstract layer in MTV paradigm.
```
artemiy312: MTV -> MVT

duker33: @artemiy312 , seems it's called `MTV`. Like one popular TV-channel

> If you’re hungry for acronyms, you might say that Django is a “MTV” framework – that is, “model”, “template”, and “view.” That breakdown makes much more sense.

https://docs.djangoproject.com/en/2.1/faq/general/#django-appears-to-be-a-mvc-framework-but-you-call-the-controller-the-view-and-the-view-the-template-how-come-you-don-t-use-the-standard-names

PR #559, opened 2018-08-31
#555 Create test_numbered_pagination_links

### LINKED ISSUE #555: tests_views.py:253-254: Create test(s) for numbered...
The puzzle `539-97439133` from #539 has to be resolved:

https://github.com/fidals/shopelectro/blob/f6ab9f062e9740d0f7114da7aa742855f30f5fbc/shopelectro/tests/tests_views.py#L253-L254

The puzzle was created by Artemiy on 29-Aug-18. 

Estimate: 30 minutes,     

artemiy312:
Closes #555

### DIFF
--- a/shopelectro/tests/tests_views.py
+++ b/shopelectro/tests/tests_views.py
@@ -252,35 +252,43 @@ class CatalogPagination(BaseCatalogTestCase):
             404,
         )
 
-    def assert_pagination_links(self, next_, prev, page_number):
+    def assert_pagination_link(self, link, page_number):
         self.assertEqual(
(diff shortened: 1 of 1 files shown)


shopelectro/tests/tests_views.py
```diff
@@ -252,35 +252,43 @@ def test_pagination_404(self):
             404,
         )
 
-    def assert_pagination_links(self, next_, prev, page_number):
+    def assert_pagination_link(self, link, page_number):
```
duker33: let's add a comment. Smth like this:
```
Page numbers from link href and from link anchor should be equal.
```

PR #582, opened 2018-09-19
#581 Get id of order's fields from backend to frontend

### LINKED ISSUE #581: Правь шаблон писем
В шаблоне письма, который приходит если оплачивать онлайн, нет строки комментариями. Как в обычном шаблоне. Добавь.

artemiy312:
Closes #581

### DIFF
--- a/front/js/components/order.es6
+++ b/front/js/components/order.es6
@@ -12,7 +12,7 @@
     remove: '.js-remove',
     paymentOptions: 'input[name=payment_type]',
     defaultPaymentOptions: 'input[for=id_payment_type_0]',
-    orderForm: $('#order-form-full').data('fields'),
+    orderFieldData: $('#order-form-full').data('fields'),
   };
 
   const config = {
@@ -63,7 +63,7 @@
     const autocompleteItem = new google.maps.places.Autocomplete(cityField, config.autocomplete);
 
     google.maps.event.addListener(autocompleteItem, 'place_changed', () => {
-      storeInput($(DOM.orderForm.city));
+      storeInput($(DOM.orderFieldData.city));
     });
   }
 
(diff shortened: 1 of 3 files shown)


front/js/components/order.es6
```diff
@@ -12,12 +12,7 @@
     remove: '.js-remove',
     paymentOptions: 'input[name=payment_type]',
     defaultPaymentOptions: 'input[for=id_payment_type_0]',
-    orderForm: {
-      name: '#id_name',
-      phone: '#id_phone',
-      email: '#id_email',
-      city: '#id_city',
-    },
+    orderForm: $('#order-form-full').data('fields'),
```
duker33: `orderFormData` might be the more good name.
We can't do any actions with this `orderForm`. We can just get it's data

front/js/components/order.es6
```diff
@@ -159,14 +154,11 @@
...
-    const orderInfo = {
-      payment_type: getSelectedPayment(),
-    };
-
-    $.each(DOM.orderForm, (name, field) => {
-      orderInfo[name] = $(field).val();
-    });
-
+    const orderInfo = Object.keys(DOM.orderForm).reduce((acc, key) => {
+      acc[key] = $(DOM.orderForm[key]).val();
```
duker33: it seems to me, that map is more simple alternative:
```
Object.keys(DOM.orderForm).reduce((key) => {
return key: $(DOM.orderForm[key]).val();
}
```

Maybe syntax is not fully working. It's my not good js

shopelectro/views/ecommerce.py
```diff
@@ -18,9 +20,13 @@ class OrderPage(ec_views.OrderPage):
     cart = SECart
 
     def get_context_data(self, request, **kwargs):
+        data = super().get_context_data(request, **kwargs)
         return {
-            **super(OrderPage, self).get_context_data(request, **kwargs),
+            **data,
             'page': CustomPage.objects.get(slug='order'),
+            'order_fields': json.dumps({
```
duker33: `order_fields_html` might be the more good name.
Usually, we send to template data structures such as arrays or dicts.
But here we have plain string to render

front/js/components/order.es6
artemiy312: `.map(...)` will produce array of objects `[{a: b}, {b: c}, ...]` instead of `{a: b, b:c, ...}`

PR #586, opened 2018-09-22
rf#169  Remove get_active method

duker33:
https://github.com/fidals/refarm-site/issues/169

### DIFF
--- a/shopelectro/views/search.py
+++ b/shopelectro/views/search.py
@@ -26,7 +26,7 @@ class Search(search_views.SearchView):
         ),
         search_engine.Search(
             name='page',
-            qs=ExcludedModelTPage.actives.all(),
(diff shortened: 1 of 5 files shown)


shopelectro/views/search.py
```diff
@@ -26,7 +26,7 @@ def get_redirect_search_entity(self):
         ),
         search_engine.Search(
             name='page',
-            qs=ExcludedModelTPage.objects.get_active(),
+            qs=ExcludedModelTPage.actives.all(),
```
artemiy312: We have to use here `ExcludedModelTPage.objects.active()`, because we override `objects` [here](https://github.com/fidals/shopelectro/blob/master/shopelectro/models.py#L222)

duker33:
this code was merged with https://github.com/fidals/shopelectro/pull/602

PR #588, opened 2018-09-24
#587 Fix a syntax error in docker-up pipe 

### LINKED ISSUE #587: .drone.yml:130-134: Fix CI's error for docker-up pipe....
The puzzle `534-ecbd61f7` from #534 has to be resolved:

https://github.com/fidals/shopelectro/blob/b15dcf014d532fb77cc8741b97eb2ec006859ee7/.drone.yml#L130-L134

The puzzle was created by Artemiy on 23-Sep-18. 

Estimate: 60 minutes,     

artemiy312:
Closes #587


duker33:
i should review it, as i see

PR #590, opened 2018-09-28
Send revenue to jsecommerce 

### LINKED ISSUE #589: Измени расчет дохода в {yandex,google}.ecommerce
есть два отчета
отчет по цели купить общая https://metrika.yandex.ru/stat/new?goal=14331608&metric=ym%3As%3Agoal%3Cgoal_id%3Ereaches&sort=-ym%3As%3Agoal%3Cgoal_id%3Ereaches&chart_type=line&period=week&table=visits&title=%D0%9F%D0%BE%D0%BA%D1%83%D0%BF%D0%BA%D0%B0&id=20644114&stateHash=5b9fafb3a76af6f1f5f22d98
отчет екоммерца по заказам 
https://metrika.yandex.ru/stat/purchase?period=week&id=20644114&stateHash=5b9fa61af33d8ae0c408b0e3
в первом 47 во втором 59 посетителей, которые что-то купили
почему ? и кто прав ? 
Может туда заходят не оплаченные заказы через кассу ? 
и хотел узнать по строке доход, в екоммерц отчете. В ней из стоимости товара вычитывается себестоимость, как и я просил ? или как ?

artemiy312:
Closes #589


shopelectro/views/ecommerce.py
```diff
@@ -54,19 +54,23 @@ class FlushCart(ec_views.FlushCart):
...
 
 class OrderSuccess(ec_views.OrderSuccess):
+
     order = Order.objects.all().prefetch_related('positions')
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        positions = context['order'].positions.all()
+        total_revenue = Product.objects.filter(
+            id__in=[p.product_id for p in positions]
```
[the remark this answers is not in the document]
artemiy312: oops

shopelectro/models.py
```diff
@@ -172,15 +162,23 @@ def payment_type_name(self):
     def set_positions(self, cart):
         """Save cart's state into Order instance."""
         self.save()
+        self.revenue = 0
         for id_, position in cart:
+            print(position['price'], position['purchase_price'])
+            print(position['quantity'])
+            import sys
+            sys.stdout.flush()
```
duker33: forgot some prints

shopelectro/models.py
```diff
@@ -172,15 +162,23 @@ def payment_type_name(self):
...
+            price_diff = position['price'] - position['purchase_price']
+            self.revenue += price_diff * position['quantity']
```
duker33: several thoughts about `revenue` field:
- it's not clear to calc revenue field in `set_positions` method
- it's not clear to keep revenue in DB. Because every order will have a state when this revenue was either set or not
- Cart is more good place to cal revenue: it focused only on positions and it's prices. Order = cart + credentials

So, i suggest to create Cart's `revenue` or `total_revenue` property: it already has total_price and total_quantity props

artemiy312: >it's not clear to calc revenue field in set_positions method

From  `set_positions` docs: `Save cart's state into Order instance.`. May be it is not the best name for this method, but it falls into the semantic bounds

>it's not clear to keep revenue in DB

We have to calculate a revenue from a cart and save it to db, because when we process a successful order, we will clear the cart, although we still need this revenue in next steps

>Cart is more good place to cal revenue

We can calculate a revenue in the cart, but we still have to save it in an order.

I will move a calculation of revenue to `Cart.revenue`

duker33: @artemiy312 

> We would put a revenue to a session

in this case `Cart.clear` method will clear the whole session instead of one revenue field.
It can be confusing state between real order sessions.

-----

As of other points: you are right here. Seems set_positions is the best place to save revenue in our arch.
But semantically it's still not the best place in the world because of semantical binding between Cart.revenue and Cart.set_positions().
It's the problem in our arch:
- now we suggest that Cart.positions == CartSession (contains all it's data). It's not true
- we should create separated CartSession and models.Cart classes
- they should have methods CartSession.as_model and models.Cart.as_session
- after `cart_session.clear()` call a binded `cart_model` should preserve it's state

Create pdd task with the link on this comment plz. We can continue discussion in it's comments if it's needed

Issue #463, opened 2018-07-30
Pin celery version to 4.1.1 and move SELENIUM_* variables to local.py/drone.py


artemiy312:
See the [issue](https://github.com/celery/kombu/issues/870) for a detail about celery.

SELENIUM_* variables are not required for prod, so let's move them from base.py

duker33:
now i pin it to `4.2.0` to prevent this prod error:
```
File "/usr/local/lib/python3.7/site-packages/kombu/utils/objects.py", line 42, in __get__
  return obj.__dict__[self.__name__]
KeyError: 'router'
```

artemiy312:
`4.2.1` does not resolve the issue. Trackeback
```python

se-python                |   File "/usr/local/lib/python3.7/site-packages/ecommerce/mailer.py", line 12, in send
se-python                |     celery_send_mail.delay(*args, **kwargs)
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/task.py", line 408, in delay
se-python                |     return self.apply_async(args, kwargs)
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/task.py", line 535, in apply_async
se-python                |     **options
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/base.py", line 708, in send_task
se-python                |     router = router or amqp.router
se-python                |   File "/usr/local/lib/python3.7/site-packages/kombu/utils/objects.py", line 44, in __get__
se-python                |     value = obj.__dict__[self.__name__] = self.__get(obj)
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/amqp.py", line 606, in router
se-python                |     return self.Router()
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/amqp.py", line 284, in Router
se-python                |     return _routes.Router(self.routes, queues or self.queues,
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/amqp.py", line 601, in routes
se-python                |     self.flush_routes()
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/amqp.py", line 289, in flush_routes
se-python                |     self._rtable = _routes.prepare(self.app.conf.task_routes)
se-python                |   File "/usr/local/lib/python3.7/site-packages/celery/app/routes.py", line 136, in prepare
se-python                |     return [expand_route(route) for route in routes]
[...]

PR #596, opened 2018-10-03
#463 Install certain celery commit

artemiy312:


artemiy312:
I am going to merge it, because now an order processing is broken

PR #613, opened 2018-10-23
#609  Apply new year styles

### LINKED ISSUE #609: Return New year style
Верни новогоднюю подложку и блок популярных товаров на главной.По вопросам -HellXi в личку

duker33:
Closes #609


duker33:
this fix is applied every year, so i merge it without review

PR #627, opened 2018-11-02
rf#197  Rm prepare_data from context

duker33:
... and customize images context

Depends on https://github.com/fidals/refarm-site/pull/206

https://github.com/fidals/refarm-site/issues/197
https://github.com/fidals/refarm-site/issues/198

### DIFF
--- a/shopelectro/context.py
+++ b/shopelectro/context.py
@@ -1,0 +1,32 @@
+import typing
+
+from catalog.context import AbstractProductsListContext
+from catalog.models import ProductQuerySet
+from images.models import Image
+
+
+class ProductImages(AbstractProductsListContext):
+
+    @property
+    def images(self) -> typing.Dict[int, Image]:
+        assert isinstance(self.products, ProductQuerySet)
+
+        images = {}
+        if self.product_pages:
+            images = Image.objects.get_main_images_by_pages(
+                self.product_pages.filter(shopelectro_product__in=self.products)
+            )
+
+        return {
+            product.id: images.get(product.page)
+            for product in self.products
+        }
+
+    def get_context_data(self):
+        return {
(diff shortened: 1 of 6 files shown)


shopelectro/views/catalog.py
```diff
@@ -254,19 +247,18 @@ def load_more(request, category_slug, offset=0, limit=0, sorting=0, tags=None):
...
+        | context.DBTemplate()  # requires TaggedCategory
     )
 
     return render(request, 'catalog/category_products.html', {
-        'products_data': context.prepare_tile_products(
-            products, models.ProductPage.objects.all()
-        ),
+        'products': products,
         'product_images': context_.get_context_data()['product_images'],
+        'product_brands': context_.get_context_data()['product_brands'],
```
artemiy312:
```suggestion
products_ctx = context_.get_context_data()
...
        'product_images': products_ctx['product_images'],
        'product_brands': products_ctx['product_brands'],
```

shopelectro/context.py
```diff
@@ -0,0 +1,32 @@
...
+
+from catalog.context import AbstractProductsListContext
+from catalog.models import ProductQuerySet
+from images.models import Image
+
+
+class ProductImages(AbstractProductsListContext):
+
+    @property
+    def images(self) -> typing.Dict[int, Image]:
```
artemiy312: it is better off being a method and not a property, because it perform heavily computations

duker33: @artemiy312 , we already have `tags`, `products`, `product_pages` in the same property-style.
Let's leave as is too keep it's style. And https://github.com/fidals/refarm-site/pull/194 with it's subtasks will solve this problem too

PR #633, opened 2018-11-15
rf#208 Remove an excess argument from get_category_descendants

artemiy312:
Related to fidals/refarm-site#208


artemiy312:
I have cancelled a build, because it require the rf 0.4.15

PR #657, opened 2018-12-04
#612  Fix inheritance

### LINKED ISSUE #612: models.py:228-236: Fix model.Manager bad inheritance Now...
The puzzle `rf169-b846c0cd` from #rf169 has to be resolved: 

https://github.com/fidals/shopelectro/blob/8a1096905808fc61125187376c4099c2ab101ec7/shopelectro/models.py#L228-L236

The puzzle was created by duker33 on 22-Oct-18. 

  

duker33:
Closes #612 

Huge amount of autochange

### DIFF
--- a/shopelectro/models.py
+++ b/shopelectro/models.py
@@ -7,6 +7,7 @@ from django.conf import settings
 from django.db import models
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
+import mptt
 
 from catalog import models as catalog_models
 from ecommerce import models as ecommerce_models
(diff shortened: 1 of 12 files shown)


shopelectro/models.py
```diff
@@ -224,28 +223,17 @@ def exclude_type(self):
...
-#  Out[2]: mptt.querysets.TreeQuerySet
-#  ```
-#  But should be `pages.models.PageQuerySet`.
-#  Or just rm all excluded staff
-#  in favor on direct excluded filter using.
-class ExcludedModelTPageManager(
-    models.Manager.from_queryset(ExcludedModelTPageQuerySet)
-):
-
+class ExcludedModelTPageManager(mptt.models.TreeManager):
```
artemiy312: You should inherit it from `pages.models.PageManger`

shopelectro/tests/tests_views.py
```diff
@@ -705,6 +705,7 @@ def test_admin_autocomplete_no_results(self):
 
     def test_search_has_no_model_pages(self):
         """Search page does not contain page with type=MODEL_TYPE and duplicated content."""
+        print(type(models.ExcludedModelTPage.objects).mro())
```
artemiy312: It seems you forgot it

PR #370, opened 2018-06-19
#367  Fix robots.txt bug

duker33:
Closes #367

PR #660, opened 2018-12-04
#370 yaml anchors

ArtemijRodionov:
Closes https://github.com/fidals/stroyprombeton/issues/370


.drone.yml
```diff
@@ -1,113 +1,98 @@
+# Anchors
+when-pr-push: &when-pr-push
+  when:
+    event: [push, pull_request]
+
+docker-compose: &docker-compose
+  <<: *when-pr-push
```
duker33: docker-compose should not contain `when-pr-push` by it's semantic

PR #663, opened 2018-12-05
#624 test-docker-compose ci pipeline

### LINKED ISSUE #624: .drone.yml:127-129: Resurrect `test-docker-compose` pipe...
The puzzle `603-db394ed9` from #603 has to be resolved: 

https://github.com/fidals/shopelectro/blob/afec9c2fadf81b58200174a1352f2fbe0c399f08/.drone.yml#L127-L129

The puzzle was created by duker33 on 01-Nov-18. 

  

artemiy312:
Closes #624

### DIFF
--- a/docker/wait-for.sh
+++ b/docker/wait-for.sh
@@ -20,6 +20,10 @@ echoerr() {
   if [ "$QUIET" -ne 1 ]; then printf "%s\n" "$*" 1>&2; fi
 }
 
+echoinfo() {
+  if [ "$QUIET" -ne 1 ]; then printf "%s\n" "$*"; fi
+}
+
 usage() {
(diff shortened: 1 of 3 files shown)


docker/wait-for.sh
```diff
@@ -20,6 +20,10 @@ echoerr() {
   if [ "$QUIET" -ne 1 ]; then printf "%s\n" "$*" 1>&2; fi
```
duker33: are we needed in strange license above?

artemiy312: >The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

It does not bother us and the license require it, so why not leave it?

PR #689, opened 2019-01-15
#683 new context

### LINKED ISSUE #683: Integrate new Context classes
You can get these classes from https://github.com/fidals/refarm-site/issues/183

artemiy312:
Closes #683

### DIFF
--- a/shopelectro/views/catalog.py
+++ b/shopelectro/views/catalog.py
@@ -31,7 +31,8 @@ def get_view_type(request):
     assert view_type in ['list', 'tile']
     return view_type
 
-# @todo 683:60m Create context class(es) for catalog page representation.
+
+# @todo #683:60m Create context class(es) for catalog page representation.
 #  Remove get_catalog_context in favor of created class(es).
 def get_catalog_context(request, category, raw_tags, page_number, per_page, sorting_index):
     all_tags = newcontext.Tags(models.Tag.objects.all())
--- a/shopelectro/views/catalog.py
+++ b/shopelectro/views/catalog.py
@@ -31,33 +31,32 @@ def get_view_type(request):
     assert view_type in ['list', 'tile']
     return view_type
 
-
+# @todo 683:60m Create context class(es) for catalog page representation.
+#  Remove get_catalog_context in favor of created class(es).
 def get_catalog_context(request, category, raw_tags, page_number, per_page, sorting_index):
     all_tags = newcontext.Tags(models.Tag.objects.all())
-
+    selected_tags = newcontext.tags.ParsedTags(
+        tags=all_tags,
+        raw_tags=raw_tags,
+    )
     if raw_tags:
-        selected_tags = newcontext.tags.Checked404Tags(
-            newcontext.tags.ParsedTags(
-                tags=all_tags,
-                raw_tags=raw_tags,
-            )
-        )
-    else:
-        selected_tags = newcontext.Tags(models.Tag.objects.none())
+        selected_tags = newcontext.tags.Checked404Tags(selected_tags)
 
+    # @todo #683:30m Remove *Tags and *Products suffixes from catalog.newcontext classes.
+    #  Rename Checked404Tags to ExistOr404.
     products = newcontext.products.OrderedProducts(
+        sorting_index=sorting_index,
         products=newcontext.products.TaggedProducts(
+            tags=selected_tags,
             products=newcontext.products.ProductsByCategory(
+                category=category,
                 products=newcontext.products.ActiveProducts(
                     newcontext.Products(
                         models.Product.objects.all(),
                     ),
                 ),
-                category=category,
             ),
-            tags=selected_tags,
         ),
-        sorting_index=sorting_index,
     )
 
     paginated_products = newcontext.products.PaginatedProducts(
@@ -71,9 +70,10 @@ def get_catalog_context(request, category, raw_tags, page_number, per_page, sort
     brands = newcontext.products.ProductBrands(paginated_products, all_tags)
     grouped_tags = newcontext.tags.GroupedTags(all_tags)
 
-    contexts = newcontext.Contexts(
-        all_tags, paginated_products, images, brands, grouped_tags,
-    )
+    contexts = newcontext.Contexts(paginated_products, images, brands, grouped_tags)
+    # @todo #683:60m Create context class(es).
+    #  Remove optional_context from get_catalog_context and
+    #  patching of page.get_template_render_context in favor of created class(es).
     optional_context = {
         'skip_canonical': selected_tags.qs().exists(),
         'total_products': products.qs().count(),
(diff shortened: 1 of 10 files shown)


shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+    view_type = request.session.get('view_type', 'tile')
+    assert view_type in ['list', 'tile']
+    return view_type
+
+
+def get_catalog_context(request, category, raw_tags, page_number, per_page, sorting_index):
+    all_tags = newcontext.Tags(models.Tag.objects.all())
+
+    if raw_tags:
+        selected_tags = newcontext.tags.Checked404Tags(
```
duker33: `Checked404Tags` looks confusing. Maybe `ExistingTags` or smth?

shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+            newcontext.tags.ParsedTags(
```
duker33: we can always pass `raw_tags` (or `uri_tags`) to Tags context.
So, ParsedTags will not be needed

shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+                tags=all_tags,
+                raw_tags=raw_tags,
+            )
+        )
+    else:
+        selected_tags = newcontext.Tags(models.Tag.objects.none())
+
+    products = newcontext.products.OrderedProducts(
+        products=newcontext.products.TaggedProducts(
```
duker33: maybe we can drop `*Products` suffix here?
We are already at the `products` module.
And using it will be clear too: `newcontext.products.Tagged` for example

shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+            products=newcontext.products.ProductsByCategory(
+                products=newcontext.products.ActiveProducts(
+                    newcontext.Products(
+                        models.Product.objects.all(),
+                    ),
+                ),
+                category=category,
+            ),
+            tags=selected_tags,
```
duker33: it'll be more clear to move short args to the first place.
This example looks more clear, i think:
```
products = context...(
    tags=my_tags_set,
    products=context...(
        products=context...(
            products=context...(
                products=context...(
            )
         )
      )
   )
```

shopelectro/views/catalog.py
```diff
@@ -171,24 +228,43 @@ class CategoryPage(catalog.CategoryPage):
...
-            | context.TaggedCategory(tags=models.Tag.objects.all())
-            | context.SortingCategory()  # requires TaggedCategory
-            | context.PaginationCategory()  # requires SortingCategory
-            | context.ProductBrands()  # requires TaggedCategory
-            | se_context.ProductImages()
-            | context.DBTemplate()  # requires TaggedCategory
+        sorting_index = int(self.kwargs.get('sorting', 0))
+
+        # @todo #683:30m Create Page context class.
+        #  Move rest of plain context data to this class.
```
duker33: it's not clear task for me. I don't understand what is `rest of plain context data`.

And `get_catalog_context` big separated method not looks good. This task request to remove it in favour of some class as i suggest.
If my suggestion is right, i agree with this thought. Context should receive this set of args in more clear way, than it does now

shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+
+    contexts = newcontext.Contexts(
+        all_tags, paginated_products, images, brands, grouped_tags,
+    )
+    optional_context = {
+        'skip_canonical': selected_tags.qs().exists(),
+        'total_products': products.qs().count(),
+        'selected_tags': selected_tags.qs(),
+    }
+    return contexts, optional_context
```
duker33: plz provide comment why you are not chaining optional_context to contexts object

shopelectro/views/catalog.py
```diff
@@ -25,6 +26,62 @@ def get_products_count(request):
...
+        selected_tags = newcontext.tags.Checked404Tags(
+            newcontext.tags.ParsedTags(
+                tags=all_tags,
+                raw_tags=raw_tags,
+            )
+        )
+    else:
+        selected_tags = newcontext.Tags(models.Tag.objects.none())
+
+    products = newcontext.products.OrderedProducts(
```
duker33: why here you are using contexts chaining throw nested calls, but below - throw Contexts args?

I mean this peace of code below
```
contexts = newcontext.Contexts(
	all_tags, paginated_products, images, brands, grouped_tags,
)
```

shopelectro/views/catalog.py
artemiy312: Is it look ok If we will rename it to `ExistOr404`?

shopelectro/views/catalog.py
artemiy312: `ParsedTags` extract tag slugs from `raw_tags` and find them in db, but it seems the `else:` branch is excess, so i will remove it

shopelectro/views/catalog.py
artemiy312: I will create todo for it

shopelectro/views/catalog.py
artemiy312: I will clarify it

shopelectro/views/catalog.py
artemiy312: It will be removed in a scope of next tasks

shopelectro/views/catalog.py
artemiy312: Because these classes are decorators and they are semantically linked to the single entity, in this case it is `Product`
 
`Contexts` just aggregate different contexts together into a single context class

shopelectro/views/catalog.py
duker33: @artemiy312 , so write it in comment plz =)

shopelectro/views/catalog.py
duker33: @artemiy312 , then `ExistingOr404`. It'll be good to have nouns in class names.
But decide it

shopelectro/views/catalog.py
```diff
@@ -71,9 +70,10 @@ def get_catalog_context(request, category, raw_tags, page_number, per_page, sort
     brands = newcontext.products.ProductBrands(paginated_products, all_tags)
     grouped_tags = newcontext.tags.GroupedTags(all_tags)
 
-    contexts = newcontext.Contexts(
-        all_tags, paginated_products, images, brands, grouped_tags,
-    )
+    contexts = newcontext.Contexts(paginated_products, images, brands, grouped_tags)
+    # @todo #683:60m Create context class(es).
```
duker33: so, now we have two subtasks here:
`Create context class(es) for catalog page representation.`
and `Create context class(es).`

Seems like duplicate.
I propose to move details about `optional_context` to the first pdd subtask body.
And remove this task

shopelectro/views/catalog.py
artemiy312: I have already created a todo with an exhaustive description for it

shopelectro/views/catalog.py
artemiy312: They are pretty different and separate. The first todo is about composition of classes for catalog views, and the second one will get rid the `optional_context` and patching of page instance

artemiy312: I will clarify these points in tasks

Issue #585, opened 2018-09-21
Не приходят оплаченные заказы на почту


themygit:
https://prnt.sc/kwtnid от 19 числа. на почту не приходил

duker33:
we have fake ids, so `8062 == 14062`

duker33:
Artemiy improved and monitored logs at https://github.com/fidals/shopelectro/issues/561
We send mails in all cases. So, the problem is in some mail side.

Currently i have no idea how we can debug it. But i'll try in some way

themygit:
Не дошел заказ 8095
и в почте 3 заказа, 4 потеряли, а метрика показывает 1 продажу

themygit:
не дошел 8103
@duker33 есть идеи почему ?
моя теория с шаблонами и некорректным заполнением - лажа ?

duker33:
it's hard task, that requires devops skills. So, `x2`

duker33:
we can test email sending with vanilla django mech:
https://docs.djangoproject.com/en/2.1/topics/testing/tools/#email-services

duker33:
we had problems with this two orders:
https://www.shopelectro.ru/admin/shopelectro/order/7833/change/?_changelist_filters=p%3D6%26q%3D
https://www.shopelectro.ru/admin/shopelectro/order/8103/change/?_changelist_filters=p%3D4%26q%3D

Both this cases went from Ya.Kassa. I'll try to cover them with tests in this task and close it

duker33:
found it in selenium section.
Task is resolved by https://github.com/fidals/shopelectro/pull/602

PR #709, opened 2019-01-26
#704 check purchase

### LINKED ISSUE #704: tasks.py:64-65: Handle errors in check_purchase. Report...
The puzzle `690-44aa8d80` from #690 has to be resolved: 

https://github.com/fidals/shopelectro/blob/ce492b12f8ee5620b12c77bd7f8809d52c91444f/shopelectro/tasks.py#L64-L65

The puzzle was created by @artemiy312 on 20-Jan-19. 

  

artemiy312:
Closes  #704

### DIFF
--- a/shopelectro/settings/base.py
+++ b/shopelectro/settings/base.py
@@ -501,3 +501,8 @@ PAGINATION_NEIGHBORS = 10
 
 PRODUCTS_ON_PAGE_PC = 48
 PRODUCTS_ON_PAGE_MOB = 12
+
+TG_BOT_TOKEN = os.environ.get('TG_BOT_TOKEN')
+TG_REPORT_ADDRESSEES = os.environ.get(
+    'TG_REPORT_ADDRESSEE', '@artemiy312,@duker33'
+).split(',')
--- a/shopelectro/tasks.py
+++ b/shopelectro/tasks.py
@@ -76,3 +76,7 @@ def check_purchase():
     order_page.load()
     order_page.fill_contacts()
     order_page.make_order()
+
+    success_page = selenium.SuccessPage(driver)
+    success_page.load()
+    assert success_page.is_success()
--- a/shopelectro/tasks.py
+++ b/shopelectro/tasks.py
@@ -2,6 +2,7 @@ from contextlib import contextmanager
 
 from django.conf import settings
 from django.core.management import call_command
+from selenium.common.exceptions import WebDriverException
 
 from shopelectro import selenium
 from shopelectro.celery import app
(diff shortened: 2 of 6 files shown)


shopelectro/settings/base.py
```diff
@@ -501,3 +501,8 @@ def get_robots_content():
 
 PRODUCTS_ON_PAGE_PC = 48
 PRODUCTS_ON_PAGE_MOB = 12
+
+TG_BOT_TOKEN = os.environ.get('TG_BOT_TOKEN')
+TG_REPORT_ADDRESSEES = os.environ.get(
+    'TG_REPORT_ADDRESSEE', '@artemiy312,@duker33'
```
duker33:
```suggestion
    'TG_REPORT_ADDRESSEES', '@artemiy312,@duker33'
```

shopelectro/tasks.py
```diff
@@ -61,18 +63,30 @@ def update_catalog():
         collect_static()
     ]
 
-# @todo #690:60m Handle errors in check_purchase.
-#  Report failed attempts. Schedule it in the celery beat.
+# @todo #690:30m Schedule check_purchase in the celery beat.
```
duker33: @artemiy312 , i already suggested asincio instead of celery beat at the parent issue's PR:
https://github.com/fidals/shopelectro/pull/695#discussion_r249236197
What do you think about it?

shopelectro/settings/base.py
duker33: let's plz create some channel "shopelectro_health" or "shopeletro_report" and send stats there

shopelectro/tasks.py
```diff
@@ -61,18 +63,30 @@ def update_catalog():
...
+        order_page = selenium.OrderPage(driver)
+        order_page.load()
+        order_page.fill_contacts()
+        order_page.make_order()
+
+        success_page = selenium.SuccessPage(driver)
+        success_page.load()
+        assert success_page.is_success()
+    except (WebDriverException, AssertionError) as err:
+        if self.request.retries + 1 > self.max_retries:
```
duker33: one failed retry seems to me enough reason to send the report. Why you choose three ones?
And we should send one report for one failing. With this code celery will spawn our telegram channel

artemiy312: >one failed retry seems to me enough reason to send the report

but what if it is just false positive temporary issue, that usually occurs.
just in case i will create a setting variable for retry number

>With this code celery will spawn our telegram channel

It's not, it will only send the message on the last attempt.

shopelectro/tasks.py
artemiy312: I believe it is bad idea:
- Django is a synchronous framework that will interfere with asynchronous code
- We need a stateful pipeline for message delivery to be able to save messages and process them later in case of deployment, bad internet connection, hardware problems, any other thing that prevent a message processing
- We already have working solution, that looks fine and aren't annoying us

artemiy312: We can move into `rq`, `huye` or something else, but it seems it will be the same picture

shopelectro/tasks.py
artemiy312: @duker33 ,
>but this last attempt occurs every max_retries check. It increases an interval of repeated messages, but still called spawn

i don't see the point. Can you describe it in more detail?
I believe it will report an error once at the end of the retry counter.
e.g. if  `max_retries = 3` then the third failed try only will be reported, and previous once will be omitted

>But this problem requires more systematic decision

Let's just set max retries number to 1 and if there are issues with this, we will find a solution

duker33: > Let's just set max retries number to 1 and if there are issues with this, we will find a solution

@artemiy312 , y, it's perfect.

About `max_retries` - merge this code plz. I'll create issue and we'll discuss it separately

shopelectro/tasks.py
duker33: i'll answer just for history:
- it's not big deal, i think. We can our own async utils process it with our own separated module
- we can provide safety by hands in every concrete case. In orders case we can (and we do) put them to the db synchronously and then do some dangerous async operations
- we had problems with celery not long time ago: unsent mails, not launched celery services and so on. Problems for example: [one](https://github.com/fidals/shopelectro/issues/585), [two](https://github.com/fidals/shopelectro/issues/561), [three](https://github.com/fidals/shopelectro/issues/424)

PR #716, opened 2019-01-30
711_fix_stale_links

### LINKED ISSUE #711: Stale links on main page
Links to category with slug `akkumulators-aaa` and maybe some similar

duker33:
Closes #711

### DIFF
--- a/shopelectro/tests/tests_views.py
+++ b/shopelectro/tests/tests_views.py
@@ -532,7 +532,7 @@ class IndexPage(TestCase):
     MAIN_PAGE_TILE = {
         'some_section': [
             {'name': 'Has url', 'url': '/section/first/'},
-            {'name': 'Has url and slug',  'slug': 'slug', 'url': '/section/second/'},
+            {'name': 'Has url and slug', 'slug': 'slug', 'url': '/section/second/'},
             {'name': 'Has slug, but not url', 'slug': 'third'},
         ]
     }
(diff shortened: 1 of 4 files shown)


shopelectro/tests/tests_views.py
```diff
@@ -527,6 +526,26 @@ def test_page_db_template_with_special_chars(self):
...
+    MAIN_PAGE_TILE = {
+        'some_section': [
+            {'name': 'Has url', 'url': '/section/first/'},
+            {'name': 'Has url and slug',  'slug': 'slug', 'url': '/section/second/'},
+            {'name': 'Has slug, but not url', 'slug': 'third'},
+        ]
+    }
+
+    def test_get_category_tile(self):
+        with override_settings(MAIN_PAGE_TILE=self.MAIN_PAGE_TILE):
```
artemiy312: You can use it as decorator

duker33: @artemiy312 , y, i can, but in this case i'll need to move `MAIN_PAGE_TILE` outside of class context

artemiy312: @duker33 `MAIN_PAGE_TILE` and `override_settings` will be in the same scope, so there won't be an issue  
```python
def a_deco(arg):
    print(arg)
    def deco_f(f):
        return f
    return deco_f
class A:
    A_VALUE = 'it will be printed'

    @a_deco(A_VALUE)
    def a(self):
        pass
```
```
it will be printed
```

duker33: @artemiy312 , y, it's true.
I'll fix it in further PRs

PR #765, opened 2019-03-12
#fix Ignore category for price

duker33:
I'll merge it without review


duker33:
failed test feels good on the local env and is not related to touched code.
Merge, prod and PO wait it

PR #774, opened 2019-03-15
#766 Move publishing of onOrderSend before submit.

### LINKED ISSUE #766: Check if Ya.Metrica is still unstable
Problem is described at #754. We installed Sentry to collect errors.
Check Sentry's front logs. Prove whether our front has stable problems with Ya.Metric or not

artemiy312:
Closes #766

### DIFF
--- a/front/js/components/order.es6
+++ b/front/js/components/order.es6
@@ -42,7 +42,6 @@
       'onCartUpdate', renderTable, fillSavedInputs,
       touchSpinReinit, restoreSelectedPayment, cityAutocomplete,
     );
-    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend'));
 
     /**
      * Bind events to parent's elements, because of dynamic elements.
@@ -217,6 +216,7 @@
         .then(formData => renderYandexForm(formData))
         .then(() => submit(DOM.yandexForm));
     } else {
+      mediator.publish('onOrderSend');
       submit(DOM.fullForm);
     }
   }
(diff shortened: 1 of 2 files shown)


front/js/components/order.es6
```diff
@@ -42,7 +42,6 @@
       'onCartUpdate', renderTable, fillSavedInputs,
       touchSpinReinit, restoreSelectedPayment, cityAutocomplete,
     );
-    $(DOM.fullForm).submit(() => mediator.publish('onOrderSend'));
```
artemiy312: The `submit` function from `submitOrder` provides a window  at 100ms for `onOrderSend` processing, but because of this line `onOrderSend` didn't use the window and executes after  

So `onOrderSend` must not subscribe on `DOM.fullForm` submit and executes before it, otherwise it won't use the window in 100ms

front/js/components/order.es6
```diff
@@ -217,6 +216,8 @@
         .then(formData => renderYandexForm(formData))
         .then(() => submit(DOM.yandexForm));
     } else {
+      // onOrderSend must be triggered before submit
+      mediator.publish('onOrderSend');
```
duker33: you reproduced the old logic here. But seems it was wrong: we should publish `onOrderSend` in isYandex case too.
Do it here or create pdd subtask plz

artemiy312: I am going to create a todo, because we should test it

PR #779, opened 2019-03-19
#fix  uncomment refarm volumes forever

duker33:
Proposal to uncomment refarm volumes.
They are always enabled as for me. And this fact creates a lot of cvs noise


artemiy312:
@duker33 Let's implement a script that will check `REFARM_DIR` existence and pull it if isn't presented. It'll be part of `deploy-dev` receipt from Makefile

duker33:
@artemiy312 , i'll create task before merge

PR #790, opened 2019-03-24
#789 get cart route

### LINKED ISSUE #789: tests_selenium.py:395-397: Load the cart lazily. It will...
The puzzle `783-dbb67fb4` from #783 has to be resolved: 

https://github.com/fidals/shopelectro/blob/63cb15c122aec4ac4992e292aefcfecf5057d0b3/shopelectro/tests/tests_selenium.py#L395-L397

The puzzle was created by Artemiy on 24-Mar-19. 

  

artemiy312:
Closes #789


shopelectro/views/ecommerce.py
```diff
@@ -34,6 +36,37 @@ def get_context_data(self, request, **kwargs):
         }
 
 
+# @todo #789:60m Make cart routes REST style instead of RPC.
+
+# @todo #789:15m Test http cache for cart-get route.
```
duker33: In this case cache should be disabled, as i understand.
Clarify it plz in subtask body

shopelectro/views/ecommerce.py
```diff
@@ -34,6 +36,37 @@ def get_context_data(self, request, **kwargs):
...
+
+
+def cart_etag(request):
```
duker33: why it's not the method of the Cart class?
The Cart class directly operates with it's session too

shopelectro/views/ecommerce.py
```diff
@@ -34,6 +36,37 @@ def get_context_data(self, request, **kwargs):
...
+    order_form = OrderForm
+
+    def get(self, request):
+        return self.json_response(request)
+
+    @classmethod
+    def as_view(cls, *args, **kwargs):
+        # Prevent django cache middleware to add default max-age
+        # this view should always revalidate a request.
+        force_revalidate = cache_control(max_age=0, must_revalidate=True)
```
duker33: it seems too complex technique with handy-mixing etags.

This approach should be much more simple as i see:
- an app gives to browser fully cached page with empty cart block
- a browser uses JS to query fresh cart data from a server
- browser renders this data at cart block

shopelectro/views/ecommerce.py
```diff
@@ -34,6 +36,37 @@ def get_context_data(self, request, **kwargs):
...
+        # Add ETag to preserve cache
+        etag_hasher = etag(cart_etag)
+        view = super().as_view(*args, **kwargs)
+        return force_revalidate(etag_hasher(view))
```
duker33: hope the manual etag will not fully change page's ETag, but only will combine with it

shopelectro/views/ecommerce.py
artemiy312: the question we should to ask is: why is the session in the Cart class. The Cart should know anything about http, cookies, session and etc.
Right now we haven't proper place for it, so i just have created this function.
I will create a todo for this

shopelectro/views/ecommerce.py
artemiy312: Yes, you are right, we shouldn't make this complicated logic of cache until it becomes an issue

shopelectro/views/ecommerce.py
artemiy312: I will just drop this with cache logic

shopelectro/views/ecommerce.py
artemiy312: I will just drop this with cache condition logic

PR #857, opened 2019-05-25
#827 product in pack

### LINKED ISSUE #827: Render package price on a product page
If product has a related tag "В упаковке" with value, render a price per pack, not price per unit.
For example for product from a screen you should change price `524 руб / шт` to `2096 руб / упак`, because this product contains four units in a package.

image

artemiy312:
Closes #827


shopelectro/management/commands/update_in_pack.py
```diff
@@ -0,0 +1,29 @@
+import logging
```
duker33: maybe separated command is not necessary?
I think it has no sense to launch pack sync with CLI

shopelectro/management/commands/update_in_pack.py
```diff
@@ -0,0 +1,29 @@
...
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from shopelectro.models import TagGroup
+
+logger = logging.getLogger(__name__)
+
+
+def update():
+    uuid = 'ae30f766-0bb8-11e6-80ea-02d2cc20e118'
```
duker33: only update_db should know this uuid as isolated place working with awful xml files.
If you move this code to update_db, the problem will gone

PR #897, opened 2019-06-24
#866 render products in pack

### LINKED ISSUE #866: Change price label for packs
Если у товара есть свойство "Упаковка" и значение больше 1, то  рядом с ценой пишем не хх руб / шт ,а хх руб / упаковка

artemiy312:
Closes #866

### DIFF
--- a/shopelectro/templatetags/se_extras.py
+++ b/shopelectro/templatetags/se_extras.py
@@ -86,7 +86,7 @@ def humanize_price(price):
 
 
 @register.filter
-def show_units(item):
+def show_price_units(item):
     if (getattr(item, 'in_pack', 1) > 1):
         return 'руб / упаковка'
     return 'руб / шт'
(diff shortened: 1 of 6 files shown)


shopelectro/templatetags/se_extras.py
```diff
@@ -85,6 +85,13 @@ def humanize_price(price):
     return intcomma(floatformat(price, 0))
 
 
+@register.filter
+def show_price_units(item):
```
duker33: We are showing a price. Units string is just a representation of this price.
```suggestion
def show_price_in_units(item: Product):
```

And one more point. It seems that a product itself could have to represent it's price. So, we should move it to Product class method.

It's up to you. But if you'll decide to preserve templates filter, tell plz why you are doing it. Your explanation will be good both for me and for project

artemiy312: >And one more point. It seems that a product itself could have to represent it's price. So, we should move it to Product class method.

I think so too, but besides Product model we have to deal with Cart's positions and Position model, so probably sooner or later we will have to share this logic for them too
Therefore I decide to do it now

PR #911, opened 2019-06-27
#910  Fix excluding products from price

### LINKED ISSUE #910: Fix products exclude from price
Production config contains `7387` product excluded, but price contains the product anyway `<offer id="7387" available="true">`

duker33:
Closes #910

### DIFF
--- a/shopelectro/models.py
+++ b/shopelectro/models.py
@@ -82,6 +82,8 @@ class Product(
         verbose_name=_('tags'),
     )
 
+    # @todo #910:30m  Explore and doc what is vendor_code.
+    #  And render it as some hidden option on the product page.
     vendor_code = models.SmallIntegerField(verbose_name=_('vendor_code'))
     uuid = models.UUIDField(default=uuid4, editable=False)
     purchase_price = models.FloatField(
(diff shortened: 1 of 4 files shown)


shopelectro/models.py
```diff
@@ -82,6 +82,8 @@ class Product(
         verbose_name=_('tags'),
     )
 
+    # @todo #910:30m  Explore and doc what is vendor_code.
+    #  And render it as some hidden option on the product page.
```
artemiy312: >render it as some hidden option

What did you mean? Should we remove this from product page?

shopelectro/models.py
```diff
@@ -82,6 +82,8 @@ class Product(
         verbose_name=_('tags'),
     )
 
+    # @todo #910:30m  Explore and doc what is vendor_code.
```
artemiy312: I still remember this story :) The root of evil here 1C integration

shopelectro/models.py
duker33: i'll just rm this string

shopelectro/models.py
duker33: @artemiy312 , then it's task for you =)

PR #917, opened 2019-06-29
#914 Document Product.vendor_code field

### LINKED ISSUE #914: models.py:85: Explore and doc what is vendor_code.
The puzzle `910-f547fee5` from #910 has to be resolved: 

https://github.com/fidals/shopelectro/blob/e7de650f01b749f584f374923d7863083cc83b3c/shopelectro/models.py#L85-L85

The puzzle was created by duker33 on 28-Jun-19. 

Estimate: 30 minutes,  role: DEV. 

artemiy312:
Closes #914

### DIFF
--- a/shopelectro/models.py
+++ b/shopelectro/models.py
@@ -82,7 +82,11 @@ class Product(
         verbose_name=_('tags'),
     )
 
-    # @todo #910:30m  Explore and doc what is vendor_code.
+    # `vendor_code` is "public id". Use this to publicly identify a product.
+    # We bring it from 1C database and sync with products.
+    # We doesn't use the id field instead, because it is auto-increment sequence,
+    # that can't be changed easily. We decided to avoid that complexity.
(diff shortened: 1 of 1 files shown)


shopelectro/models.py
```diff
@@ -82,7 +82,11 @@ class Product(
         verbose_name=_('tags'),
     )
 
-    # @todo #910:30m  Explore and doc what is vendor_code.
+    # `vendor_code` is "public id". Use this to publicly identify a product.
+    # We bring it from 1C database and sync with products.
```
duker33: i would like to underline that `vendor_code` comes from SE stock.
Maybe you'll include these words to the comment:
```
`vendor_code` is a unique id for stock keeping unit.
It differs from site's product id and we use it to expose to the customer:
we render it at site pages, market yml offers and so on.
```
stock keeping unit (SKU) is a common term for stocked products: https://en.wikipedia.org/wiki/Stock_keeping_unit

PR #922, opened 2019-07-01
#903  Create slider for mobile version

### LINKED ISSUE #903: main.less:449: Markup mobile tile-about and text for main...
The puzzle `891-9180fe3b` from #891 has to be resolved: 

https://github.com/fidals/shopelectro/blob/bc1d3db1da1faa557d353c2ff9ac4dbfe9296c5c/front/less/pages/main.less#L449-L449

The puzzle was created by @duker33 on 26-Jun-19. 

Estimate: 30 minutes,  role: DEV. 

duker33:
Closes #903

image

### DIFF
--- a/front/js/components/main.es6
+++ b/front/js/components/main.es6
@@ -19,19 +19,12 @@
   };
 
   const init = () => {
-    if ($(window).width() < SCREENS.sm) {
-      $('.tile-about .row').slick({
-        dots: true,
-        arrows: true,
-        mobileFirst: true,
-      });
-    }
-
     fillInUserData({
       phone: localStorage.getItem(configs.labels.phone),
       time: localStorage.getItem(configs.labels.callTime),
     });
     setUpListeners();
+    initSlider();
   };
 
   /**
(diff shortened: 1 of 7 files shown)


front/js/components/main.es6
```diff
@@ -8,12 +8,23 @@
     $timeTag: $('.js-select-time'),
   };
 
+  // @todo #903:30m  Move SCREENS const to configs.es6 file.
```
artemiy312:
```suggestion
  // @todo #903:15m  Move SCREENS const to configs.es6 file.
```

It is only about moving the single variable

front/js/components/main.es6
```diff
@@ -38,6 +49,16 @@
     );
   }
 
+  function initSlider() {
+    if ($(window).width() < SCREENS.sm) {
+      $('.tile-about .row').slick({
```
artemiy312: We should check what will do the library on window resizing and probably make a handler for transitions between sizes (e.g. `sm -> lg`)

artemiy312:
Let's make a test for it

front/js/components/main.es6
duker33: i'll subtask it

PR #925, opened 2019-07-02
#880 Create CatalogBlock model 

### LINKED ISSUE #880: catalog.py:45-47: Improve categories matrix arch. Now...
The puzzle `837-edb24a44` from #837 has to be resolved: 

https://github.com/fidals/shopelectro/blob/bd9b654124b09163e5f1bb3a77f467fff56cffcd/shopelectro/views/catalog.py#L45-L47

The puzzle was created by duker33 on 17-Jun-19. 

Estimate: 60 minutes,  role: DEV. 

artemiy312:
Closes #880


shopelectro/tests/tests_models.py
```diff
@@ -86,3 +86,32 @@ def test_get_brands_from_products(self):
                 p: p.tags.filter(group__name=settings.BRAND_TAG_GROUP_NAME).first()
                 for p in products
             }
+
+    def test_get_catalog_blocks(self):
+        """Perform four queries to fetch in batch blocks, categories, pages and category's children."""
+        roots = Category.objects.filter(level=0)
+        roots_count = roots.count()
```
duker33: we can use `roots.count()` directly, without var

shopelectro/tests/tests_models.py
```diff
@@ -86,3 +86,32 @@ def test_get_brands_from_products(self):
                 p: p.tags.filter(group__name=settings.BRAND_TAG_GROUP_NAME).first()
                 for p in products
             }
+
+    def test_get_catalog_blocks(self):
```
duker33:
```suggestion
    def test_get_catalog_blocks_queries(self):
```

The name will able us to get rid from pydoc. It just doubles code lines now

shopelectro/tests/tests_models.py
```diff
@@ -86,3 +86,32 @@ def test_get_brands_from_products(self):
...
+        """Perform four queries to fetch in batch blocks, categories, pages and category's children."""
+        roots = Category.objects.filter(level=0)
+        roots_count = roots.count()
+        for category in roots:
+            CatalogBlock.objects.create(category=category)
+
+        with self.assertNumQueries(4):
```
duker33: This structure would be more clear. Up to you

```
with self.assertNumQueries(1):
    # fetch blocks
with self.assertNumQueries(3):
    # fetch category, page, rows
```

shopelectro/tests/tests_models.py
```diff
@@ -86,3 +86,32 @@ def test_get_brands_from_products(self):
...
+class CatalogBlockModel(TestCase):
+
+    fixtures = ['dump.json']
+
+    def test_rows_count(self):
+        first, second, *_  = Category.objects.all()
+        block = CatalogBlock.objects.create(category=first)
+        sized_block = CatalogBlock.objects.create(category=second, block_size=2)
+
+        self.assertNotEquals(first.children.active(), len(sized_block.rows()))
```
duker33: you are comparing QS with number here. Let me suggest what you meaned

```suggestion
        self.assertEquals(first.children.active().count(), sized_block.rows().count())
```

duker33:
@artemiy312 , i guess your code contains too many work for 1h. I recommend you to increase task price to 2h

shopelectro/tests/tests_models.py
artemiy312: It'll bring an additional query to `assertNumQueries(4)` scope, that doesn't related to the test

shopelectro/tests/tests_models.py
artemiy312: It is already in a relevant class `QueryQuantities`, so I'll remove the comment

shopelectro/tests/tests_models.py
artemiy312: Looks better, but `blocks` can't be fetched in 1 query, because of `select_related('category', 'cateogyr__page')` and `rows` takes 2 queries, so this structure be more relevant:
``` 
with self.assertNumQueries(2):
    # fetch blocks, category, page
with self.assertNumQueries(2):
    # fetch rows
```
P.S.: [To make clear](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#prefetch-related) the point about `select_related`
>select_related gets the related objects in the same database query

shopelectro/tests/tests_models.py
artemiy312: Sorry for the mess. I have reworked the test

PR #926, opened 2019-07-02
#920 #921 test js ecommerce

### LINKED ISSUE #920: Test product add/remove Ya.ecommerce goals.
The puzzle `808-b0743766` from #808 has to be resolved: 

https://github.com/fidals/shopelectro/blob/5db3a865dfff6ad0d4de0ff7c94b072713d64a91/shopelectro/tests/tests_js_analytics.py#L90-L93

The puzzle was created by Artemiy on 30-Jun-19. 

Estimate: 120 minutes,  role: DEV. 

artemiy312:
Closes #920 #921


shopelectro/selenium/elements/product.py
```diff
@@ -1,16 +1,21 @@
...
-from selenium.webdriver.common.by import By
-
 
 class Product(abc.ABC):
 
     def name(self):
         raise Unavailable('determine the product name.')
 
+    def vendor_code(self):
+        raise Unavailable('determine the product name.')
```
artemiy312: TODO

shopelectro/selenium/pages/category.py
```diff
@@ -18,11 +20,37 @@ def __init__(self, driver, slug):
...
-    def product_cards(self) -> typing.List[CatalogCard]:
-        raise NotImplementedError
+    def product_cards(self) -> typing.List[elements.CatalogCard]:
+        products_count = len(self.driver.find_elements(
+            By.CLASS_NAME, 'product-card'
+        ))
+        return [elements.CatalogCard(self.driver, index=i) for i in range(products_count)]
+
+    def find_card(self, id_: int) -> elements.CatalogCard:
+        return elements.CatalogCard(self.driver, id_=id_)
```
artemiy312: >id_=id_

Right now we have exposed `Product.id` field in some places to communicate with front.
@duker33  Should we remove it in favor  `Product.vendor_code`? I'd prefer to remove `Product.id` from front totally. It'll get rid of ambiguous for us

shopelectro/selenium/elements/product.py
```diff
@@ -26,17 +31,43 @@ def remove_from_cart(self):
...
+        product_xpath = '//*[@id="products-wrapper"]'
+
+        if self._id:
+            return f'{product_xpath}//*[@data-product-id="{self._id}"]/{path}'
+
+        # xpath indexes starts from 1
+        return f'{product_xpath}/div[{self._index + 1}]/{path}'
+
+    def vendor_code(self):
+        if self._id:
```
artemiy312: I'll remove this predicate, because `self._id != vendor_code`. I made this because I forget that we are still using Product.id in front

shopelectro/selenium/elements/cart.py
```diff
@@ -22,20 +25,33 @@ def _hover(self):
             (By.CLASS_NAME, 'js-cart-wrapper')
         ))
 
+    @contextmanager
+    def wait_changes(self):
+        def wait_changes(browser):
+            try:
+                return positions_before != self.positions()
```
duker33: try-except looks dangerous. Lot's look at the cart clearing case:
- we wrap non-empty cart clearing with this context manager
- after clearing positions_before is not empty yet
- but self.positions() are already empty
- so `positions_before != self.positions()` should be `True`
- but with this try-except wrapping the function will return `False`

Try to wrap self..clear() with the context manager to check my assumption.

You should catch the exception only for `self.positions()`, not the whole expression `return positions_before != self.positions()`.

shopelectro/selenium/elements/cart.py
```diff
@@ -22,20 +25,33 @@ def _hover(self):
...
-        positions_count = len(self.driver.find_elements(
-            By.CLASS_NAME, 'basket-item'
-        ))
+        try:
+            # use short_wait to avoid long pauses in case of the empty cart
+            positions_count = len(self.driver.short_wait.until(EC.presence_of_all_elements_located(
+                (By.CLASS_NAME, 'basket-item')
+            )))
+        except TimeoutException:
+            return []
```
duker33:
```suggestion
            positions_count = 0
```

it would be more idiomatic, but it's up to you

shopelectro/selenium/elements/product.py
```diff
@@ -26,17 +31,43 @@ def remove_from_cart(self):
...
         """
         Ctor.
 
-        :param int card_index: The index number of the product card at a category page
+        :param int index: The index number of the product card at a category page
         """
         self.driver = driver
-        self.xpath = f'//*[@id="products-wrapper"]/div[{card_index}]/div[2]/div[5]/'
+
+        if (index is None and not id_) or (index and id_):
```
duker33: is semantic coupling between arguments. It would be more good to create two separated constructors: one for `index` and another for `id_`.

Python doesn't allow us to create multiple constructors, but we can use static class methods instead:

```
class CatalogCard(Product):

    def __init__(self, index: int, *args, **kwargs):
        ...

    @classmethod
    def with_id(cls, id_, *args, **kwargs):
        return cls(...)
```

shopelectro/selenium/pages/category.py
duker33: @artemiy312 , i thought about "id vs vendor_code" problem. I would prefer to save `id`. Let me explain my position.

`vendor_code` comes to us from stock. It should be called `stock_id` in fact. It comes from outside, we don't control it. Tomorrow it can turn from int to str or become not unique or smth. So, we should use it only as an informational code for our users. And should not rely on it to build our infrastructure.

What we need now is responsibility separation between `id` and `vendor_code` throw it interfaces

shopelectro/selenium/pages/category.py
```diff
@@ -18,11 +20,37 @@ def __init__(self, driver, slug):
...
 
-    def add_to_cart(self, products: typing.List[CatalogCard] = None):
-        default = [CatalogCard(self.driver, i) for i in range(1, 7)]
+    def load_more(self):
+        old_len = len(self.product_cards())
+        load_more_locator = (By.ID, 'btn-load-products')
```
duker33:
```suggestion
        locator = (By.ID, 'btn-load-products')
```

we are already in `load_more` method. Up to you

shopelectro/selenium/pages/category.py
artemiy312: @duker33 I'll create an issue for this discuss

shopelectro/selenium/elements/cart.py
artemiy312: >so positions_before != self.positions() should be True
but with this try-except wrapping the function will return False

`self.positions()` checks cart emptiness, so it dosn't need this try/catch, but it is needed for cart's position hash calculation.  

I'll create todo to clarify the point

>Try to wrap self..clear() with the context manager to check my assumption.

We already have a `remove` method that behaves the same when positions quantity = 1.  We already have used it in tests

PR #945, opened 2019-07-16
#fix  Rm backup docker service

duker33:
Current backup system takes too much disk space on serv.
Our files and DB are not changed too frequently. So it should do incremental backups instead of the full ones every time.

Mentioned incremental backups idea at the https://github.com/fidals/stroyprombeton/issues/188
We'll implement them with this task or it's subtask

### DIFF
--- a/docker/docker-compose-production.yml
+++ b/docker/docker-compose-production.yml
@@ -143,23 +143,6 @@ services:
     volumes:
       - /dev/shm:/dev/shm
 
-  # Контейнер делает бэкапы по крону.
-  # Бэкап - это несколько архивов: database.tar.gz, media.tar.gz, static.tar.gz
-  # в директории /opt/backup/shopelectro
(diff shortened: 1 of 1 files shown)


docker/docker-compose-production.yml
```diff
@@ -143,23 +143,6 @@ services:
     volumes:
       - /dev/shm:/dev/shm
 
-  # Контейнер делает бэкапы по крону.
-  # Бэкап - это несколько архивов: database.tar.gz, media.tar.gz, static.tar.gz
-  # в директории /opt/backup/shopelectro
-  backup-data:
```
artemiy312: @duker33 Please, provide a comment why do you remove it

duker33: @artemiy312 , sorry. Leaved explanation at the PR body

PR #965, opened 2019-09-21
#758  Add custom 500 error page

### LINKED ISSUE #758: Create a custom page for 500 error
The page should have an apology and an suggestion to continue the purchase with the help of a consultant

duker33:
Closes #758 

image

### DIFF
--- a/shopelectro/urls.py
+++ b/shopelectro/urls.py
@@ -1,15 +1,15 @@
-from datetime import timedelta
 from collections import OrderedDict
+from datetime import timedelta
 
+from django import views as dj_views
 from django.conf import settings
 from django.conf.urls import url, include
 from django.conf.urls.static import static
 from django.contrib.sitemaps.views import sitemap
 from django.views.decorators.cache import cache_page, never_cache
 
-from pages.views import RobotsView, SitemapPage
 from pages.urls import custom_page_url
-
+from pages.views import RobotsView, SitemapPage
 from shopelectro import sitemaps, views
 from shopelectro.admin import se_admin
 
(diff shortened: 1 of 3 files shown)


shopelectro/urls.py
```diff
@@ -1,15 +1,15 @@
-from datetime import timedelta
 from collections import OrderedDict
+from datetime import timedelta
 
+from django import views as dj_views
```
artemiy312: `TemplateView` should be enough, but up to you

shopelectro/urls.py
```diff
@@ -132,3 +132,12 @@ def cache_page(arg):  # Ignore PyFlakesBear
         *static(settings.STATIC_URL, document_root=settings.STATIC_ROOT),
         *static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT),
     ]
+
+if settings.TEST_ENV or settings.DEBUG:
```
artemiy312: Why did you declare 404 and 500 only for these environments?  We need them in production too, right?
```suggestion
if settings.TEST_ENV or not settings.DEBUG:
```

duker33: @artemiy312 , noup. We need them only for dev and test envs.
This if is needed because Django hides real 404 and 500 pages on dev and test env to show traceback. So, i added them manually to for testing at test env and manual checking at dev env.
I'll add comment about it

artemiy312: @duker33 500 page contains a lot of useful debugging info. I'd rather enable these views only for settings.TEST_ENV. What do you think?

duker33: @artemiy312 , sorry, didn't noticed your comment before merge. Anyway, let's continue discussion there.

I don't disable debugging info for 500 errors. I just add new `/500` url that responses with real production version of 500 error page and 200 ok status

PR #987, opened 2019-10-01
#986  First leveled children

### LINKED ISSUE #986: Render only first-leveled children at the header menu
Now we have too many categories at the header menu
image

duker33:
Closes #986


duker33:
i'll merge it directly after CI will pass. Prod waits fix

PR #995, opened 2019-10-09
#991  Fix stuck header menu tests

### LINKED ISSUE #991: Resolve stuck tests
CI fails because of stuck tests. They are working at the local and relevant code looks like they should pass
https://ci.fidals.com/fidals/shopelectro/1727/9

duker33:
Closes #991

### DIFF
--- a/shopelectro/settings/drone.py
+++ b/shopelectro/settings/drone.py
@@ -5,6 +5,9 @@ from .base import *
 
 DEBUG = True
 
+# Header categories menu uses cache in templates.
+# Disable cache to avoid stale menu testing.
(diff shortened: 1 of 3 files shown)


shopelectro/settings/drone.py
```diff
@@ -5,6 +5,12 @@
 
 DEBUG = True
 
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
```
artemiy312: Please, provide a comment for this. Why do we need this backend? How does it relate to header tests?

PR #1012, opened 2019-11-01
Avoided hardcoded values in RSS template

olegush:
Closes #1009

PR #1019, opened 2019-11-18
#1012 Avoided hardcoded values in RSS template 

olegush:
Closes #1012


duker33:
@olegush , see how PRs body should look like. I've fixed it
image

olegush:
@duker33 , got it, thanks!

